PR #34, opened 2016-03-14
Use of client interface

KtorZ:
Hey!
It wasn't possible to mock the client interface without rewriting it, the Subscribe* methods used a `MessageHandler` which was an interface using a pointer to the client `Client` struct.

`type MessageHandler func(*Client, Message)`

To enable a nice testing, I removed the `Client` struct from the exported entities and only exported the `ClientInt` interface (renamed to `Client`) such that we only interact with the interface itself, rather than with the struct which was only carrying unexported fields. So, the change might be transparent for users, but it now allows an easier testing for software using the paho client :) 

Cheers.

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,20 +34,10 @@ const (
 	connected
 )
 
-// ClientInt is the interface definition for a Client as used by this
+// Client is the interface definition for a Client as used by this
 // library, the interface is primarily to allow mocking tests.
-type ClientInt interface {
-	IsConnected() bool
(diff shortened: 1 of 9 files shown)


alsm:
Hi, thanks for your work on this. Eclipse projects require that contributors sign the CLA. There is information on this in the "Details" link against the check that failed above. Hopefully you are able to complete the process, please let me know if there are any problems.

PR #36, opened 2016-03-14
Use of client interface

KtorZ:
This time (see #34), with certificates and signatures (hopefully)

:)

### DIFF
--- a/client.go
+++ b/client.go
@@ -57,7 +57,6 @@ type Client interface {
 	IsConnected() bool
 	Connect() Token
 	Disconnect(uint)
-	disconnect()
 	Publish(string, byte, bool, interface{}) Token
 	Subscribe(string, byte, MessageHandler) Token
 	SubscribeMultiple(map[string]byte, MessageHandler) Token
--- a/client.go
+++ b/client.go
@@ -34,20 +34,10 @@ const (
 	connected
 )
 
-// ClientInt is the interface definition for a Client as used by this
+// Client is the interface definition for a Client as used by this
 // library, the interface is primarily to allow mocking tests.
-type ClientInt interface {
-	IsConnected() bool
-	Connect() Token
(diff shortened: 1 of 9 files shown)


alsm:
Thanks for doing the CLA, I like the changes, and we aren't 1.0 yet, so not bound to keep the API stable, but if you go into the samples directory and try and build them this breaks them all as the signature for a MessageHandler has changed. Similarly this will break everyone who uses the library and defines a MessageHandler.
I'll probably merge the change in a few days after I've posted to the Paho dev mailing list and a couple of other places to give some heads up that it's coming. I think the proper mocking of the interface for testing is worth it.

client.go
```diff
@@ -62,7 +52,20 @@ type ClientInt interface {
 // information can be found in their respective documentation.
 // Numerous connection options may be specified by configuring a
 // and then supplying a ClientOptions type.
-type Client struct {
+
+type Client interface {
+	IsConnected() bool
+	Connect() Token
+	Disconnect(uint)
+	disconnect()
```
brycereitano: A small comment here, having a private method in an interface makes it impossible for external packages to modify the implementation of that method. Which would be beneficial in testing.

KtorZ: The method was already there in the previous interface. I think there shouldn't be any issue to move that method from the interface to the struct itself.

brycereitano: That would be all that would need to be changed. I'm also not a maintainer of this project, so take my comment with a grain of salt. Thanks for the reply!

alsm: Yeah the lowercase disconnect() is an internal function of the implementation rather than the interface. If we can remove it before we merge this.

KtorZ:
I'll open a new one. I've removed some commits that pursue another goal (fixing dataraces). I'll make another appropriated PR later to target this issue.

PR #38, opened 2016-03-23
Enable the Store functionality

### LINKED ISSUE #37: Store stuff doesn't actually work
Looks like the code to persist pending outbound/inbound messages was [commented out at some point](https://github.com/eclipse/paho.mqtt.golang/blob/183329365eb3c151b3f8403f90c89cf4e21ed739/net.go#L116). It would be bring this feature back, or remove it entirely if it's not going to be supported in a future release. At the moment it's quite confusing that I'm calling `opts.SetStore(mqtt.NewFileStore("./some_directory"))` but not getting anything stored there!

isaacseymour:
- persist outbound messages before they get enqueued to be sent
- persist inbound messages as soon as they're received

I believe this closes #37, but I'm probably missing something. I'm also not sure where to write tests for this (I'm pretty new to Go, sorry!).

@alsm does this look ok?

### DIFF
--- a/client.go
+++ b/client.go
@@ -463,6 +463,11 @@ func (c *client) Publish(topic string, qos byte, retained bool, payload interfac
 	}
 
 	DEBUG.Println(CLI, "sending publish message, topic:", topic)
+	if pub.Qos != 0 && pub.MessageID == 0 {
+		pub.MessageID = c.getID(token)
+		token.messageID = pub.MessageID
+	}
+	persistOutbound(c.persist, pub)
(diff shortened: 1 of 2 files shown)


alsm:
Yeah, this should fix up the immediate issue of the persist code not getting used at all, although it's been a while since all of that was touched so it should probably get a closer look and some more testing to make sure it's working as it's supposed to.
Also lines 111-115 in net.go are now redundant as only Publish() pushes things on the obound channel and it will have already have set the message ID

isaacseymour:
Good point - updated.

PR #39, opened 2016-03-25
Use of client interface

KtorZ:


alsm:
Thanks again for this :+1:

Issue #32, opened 2016-02-24
client loses connection to broker (on Raspberry Pi 2)


jinie:
I'm seeing a weird issue with the client, which only happens when running on a Raspberry Pi 2.

```
ratatosk ➜  ~  go version
go version go1.3.3 linux/arm
```

I'm using the following code to connect to the broker
```go
url := fmt.Sprintf("tcp://%s:%d", host, port)
	opts := MQTT.NewClientOptions().AddBroker(url)
	opts.SetDefaultPublishHandler(OnMessage)
	opts.SetConnectionLostHandler(ConnectionLost)

	logger.Debug("Connecting to ", url)
	c := MQTT.NewClient(opts)
	if token := c.Connect(); token.Wait() && token.Error() != nil {
		panic(token.Error())
	}
	logger.Debug("Subscribing to ", topic)
	//subscribe to the topic /go-mqtt/sample and request messages to be delivered
	//at a maximum qos of zero, wait for the receipt to confirm the subscription
	if token := c.Subscribe(topic, 2, nil); token.Wait() && token.Error() != nil {
		fmt.Println(token.Error())
		os.Exit(1)
	}
```

And when running on my MacBook, it runs just fine, but when running on my RPi2 it will disconnect from the broker every 30-60 seconds, and while it seems to reconnect (per the logs), it never receives anything, it simply just "hangs".

Here's a debug trace 
```
[net]      Received Message
[net]      logic got msg on ibound
[net]      received pubrel, id: 33
[net]      logic waiting for msg on ibound
[net]      obound priority msg to write, type *packets.PubcompPacket
[net]      outgoing waiting for an outbound message
13:58:49.209 UpdateLastUp ▶ DEBUG  Updating last update for  fenrir : 28-0014550
75dff : 2016-02-24 13:58:49.110356 +0000 UTC
[pinger]   keepalive sending ping
[net]      Received Message
[net]      logic got msg on ibound
[net]      received pingresp
[net]      logic waiting for msg on ibound
[net]      outgoing stopped
[net]      incoming stopped
13:59:29.146 ConnectionLo ▶ ERROR  Connection Lost: pingresp not received, disconnecting
[client]   enter reconnect
[client]   about to write new connect msg
[client]   socket connected to broker
[client]   Using MQTT 3.1.1 protocol
[net]      connect started
[net]      received connack
[client]   client is reconnected
[net]      outgoing started
[net]      outgoing waiting for an outbound message
[net]      logic started
[...]

PR #42, opened 2016-04-03
Fix ping test (make it fail on conn error)

brocaar:
In this pull-request are two things:

* I made the ``FVTAddr`` configurable by using an optional environment variable ``TEST_FVT_ADDR``
* I updated the ping test so that it actually fails because of the issue described in #32 and https://github.com/brocaar/loraserver/issues/5

### DIFF
--- a/fvt_client_test.go
+++ b/fvt_client_test.go
@@ -987,22 +987,21 @@ func Test_MultipleURLs(t *testing.T) {
 	c.Disconnect(250)
 }
 
-/*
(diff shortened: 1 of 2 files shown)


brocaar:
I noticed that in https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L182 ``pingTimer`` is set to ``PingTimeout``. Should this have been ``KeepAlive``?

I changed it locally and ran the tests again, however, it is still failing.

PR #44, opened 2016-04-07
protect from double close of stop chan

koron:
Paho client will close `stop` channel twice,
when calling `Disconnect()` and server shutdown at same time.


yongpingchen:
+1

koron:
thanks to merge!

PR #45, opened 2016-04-13
Make reading and writing to FileStore more reliable

isaacseymour:
- use a write-ahead strategy instead of backup (so *.msg should always
  contain an uncorrupted message)
- backup, log, and ignore unreadable messages from the FileStore, rather
  than panicking

Encountered an issue with a `.msg` file getting only partially written, which meant our app could no longer run (it would panic whenever it tried to read the corrupted file, as `FileStore.restore` was never being run to restore from the backup)...

By writing to a temporary file and then doing an atomic rename* this should make it harder to end up with a corrupt file in the first place. If that ever does happen, I've updated `FileStore.Get` to log a warning and archive the file to have a `.CORRUPT` extension, for debugging.

@alsm does this look okay?

### DIFF
--- a/filestore.go
+++ b/filestore.go
@@ -15,7 +15,6 @@
 package mqtt
 
 import (
-	"io"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


alsm:
I'm happy for you to go wild with the FileStore :) It needs some looking at, please feel free to take it apart completely and redo it if you have better ideas for how it should operate.

Issue #50, opened 2016-06-01
Clients on localhost constantly disconnect/reconnect to broker


tgkavanagh:
We have been having issues where clients on the local host would disconnect from the broker due to a ping response timeout even though the response was received.  Once the problem starts, it progressively gets worse.  

**2016/05/27 22:00:53 ping.go:33: [pinger]   keepalive sending ping**
2016/05/27 22:00:53 net.go:81: [net]      Received Message
2016/05/27 22:00:53 net.go:176: [net]      logic got msg on ibound
**2016/05/27 22:00:53 net.go:180: [net]      received pingresp**
2016/05/27 22:00:53 net.go:172: [net]      logic waiting for msg on ibound
**2016/05/27 22:00:53 ping.go:40: [pinger]   pingresp not received, disconnecting**
2016/05/27 22:00:53 net.go:108: [net]      outgoing stopped
2016/05/27 22:00:53 net.go:267: [net]      logic stopped
2016/05/27 22:00:53 net.go:88: [net]      incoming stopped
2016/05/27 22:00:53 client.go:264: [client]   enter reconnect
2016/05/27 22:00:53 client.go:275: [client]   about to write new connect msg
2016/05/27 22:00:53 client.go:553: Connection lost: pingresp not received, disconnecting
2016/05/27 22:00:53 client.go:278: [client]   socket connected to broker
2016/05/27 22:00:53 client.go:285: [client]   Using MQTT 3.1.1 protocol
2016/05/27 22:00:53 client.go:349: [net]      connect started
2016/05/27 22:00:53 client.go:367: [net]      received connack
2016/05/27 22:00:53 net.go:102: [net]      outgoing started
2016/05/27 22:00:53 net.go:105: [net]      outgoing waiting for an outbound message
2016/05/27 22:00:53 net.go:169: [net]      logic started
2016/05/27 22:00:53 net.go:172: [net]      logic waiting for msg on ibound
2016/05/27 22:00:53 client.go:331: [client]   client is reconnected
2016/05/27 22:00:53 net.go:75: [net]      incoming started
2016/05/27 22:00:53 ping.go:24: [pinger]   keepalive starting

PR #51, opened 2016-06-01
Fix Ping Response Timer race condition (Issue #50)

tgkavanagh:
Use a channel to send the ping response event to the keepalive
loop and have it stop the PingRespTimer within the same context
as that used to reset it.  This prevents the timer race condition
that exists with clients on the localhost and stops local clients
from constantly disconnecting/reconnecting..


alsm:
Thanks for this, I changed it slightly to use a chan struct{} as we don't really care about the value coming down the channel.

PR #58, opened 2016-08-25
check decodeLength error

jl2005:
`decodeLength` can't return error, when connection closed. It will cause a panic in the `Unpack`.

```
func (p *PublishPacket) Unpack(b io.Reader) {
    var payloadLength = p.FixedHeader.RemainingLength
    p.TopicName = decodeString(b)
    if p.Qos > 0 {
        p.MessageID = decodeUint16(b)
        payloadLength -= len(p.TopicName) + 4
    } else {
        payloadLength -= len(p.TopicName) + 2
    }
    p.Payload = make([]byte, payloadLength)
    b.Read(p.Payload)
}
```

`payloadLength` will be  a negative number

### DIFF
--- a/packets/packets.go
+++ b/packets/packets.go
@@ -106,7 +106,10 @@ func ReadPacket(r io.Reader) (cp ControlPacket, err error) {
 	if err != nil {
 		return nil, err
 	}
-	fh.unpack(b[0], r)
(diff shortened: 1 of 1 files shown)


ddatsh:
+1

PR #61, opened 2016-10-04
Handle ping timers in single goroutine.

brycereitano:
The changes here are to stop a race condition when reconnecting. In the process, enforcing that access to the timers only happen within the goroutine. In the process, I decided to make sure that there wasn't a race condition within it's own timer. As specified in the Golang docs. https://golang.org/pkg/time/#Timer.Reset


alsm:
Great, thanks.

PR #78, opened 2017-02-15
Send origin header matching host to support cross-origin ws security

jeremymailen:
**Why**
To support websocket servers which may implement this security recommendation:
https://en.wikipedia.org/wiki/Same-origin_policy#WebSockets

A more complete solution would involve creating an option in the Paho MQTT client for specifying a customer origin value.
The behavior in this commit would be the default if unspecified.
https://devcenter.heroku.com/articles/websocket-security#origin-header

**What**
- when connecting on ws:// or wss:// simulate the origin value a browser would send on the matching http origin.

### DIFF
--- a/net.go
+++ b/net.go
@@ -17,6 +17,7 @@ package mqtt
 import (
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"net"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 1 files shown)


jeremymailen:
Bump. I have signed the ECA now.

alsm:
I have confirmed the signed CLA, merging. Thanks.

Issue #83, opened 2017-03-21
occasional deadlock when connections are lost.


databus23:
Every now and then I'm having clients getting stuck with something that seems like an logic deadlock.
The clients are running with critical and error logging (so mqtt warning and debug logs are not printed).
The only message I see in the logs is
```
MQTT CRITICAL [pinger]   pingresp not received, disconnecting
```
After this message nothing seems to happens anymore.

I finally managed to take a stckdump and this is what the relevant goroutines look like. From the timing perspective this matches closely with the above error message: The error was logged about 108 minutes before I took the stackdump. 
```
goroutine 40 [semacquire, 108 minutes]:
sync.runtime_notifyListWait(0xc4202b1240, 0xc400000001)
	/usr/local/go/src/runtime/sema.go:297 +0x10b
sync.(*Cond).Wait(0xc4202b1230)
	/usr/local/go/src/sync/cond.go:57 +0x89
io.(*pipe).read(0xc4202b1200, 0xc4202d603e, 0xfc2, 0xfc2, 0x0, 0x0, 0x0)
	/usr/local/go/src/io/pipe.go:47 +0x104
io.(*PipeReader).Read(0xc42016e0e0, 0xc4202d603e, 0xfc2, 0xfc2, 0xc42005de58, 0x6ccd3a, 0xc42020d040)
	/usr/local/go/src/io/pipe.go:130 +0x4c
bufio.(*Scanner).Scan(0xc42005df40, 0xc42019bf30)
	/usr/local/go/src/bufio/scan.go:207 +0x294
github.wdf.sap.corp/monsoon/arc/vendor/github.com/Sirupsen/logrus.(*Logger).writerScanner(0xc42020d040, 0xc42016e0e0)
	/tmp/build/80754af9/gopath/src/github.wdf.sap.corp/monsoon/arc/vendor/github.com/Sirupsen/logrus/writer.go:20 +0x86
created by github.wdf.sap.corp/monsoon/arc/vendor/github.com/Sirupsen/logrus.(*Logger).Writer
	/tmp/build/80754af9/gopath/src/github.wdf.sap.corp/monsoon/arc/vendor/github.com/Sirupsen/logrus/writer.go:12 +0x107

goroutine 51 [semacquire, 108 minutes]:
sync.runtime_Semacquire(0xc42018549c)
	/usr/local/go/src/runtime/sema.go:47 +0x34
sync.(*WaitGroup).Wait(0xc420185490)
[...]

PR #84, opened 2017-03-22
Avoid race during shutdown that can deadlock

databus23:
This is a proposed fix for #83.
Its not super elegant as it drops a message that was received over the wire but its a simple solution that avoids a deadlock.

I'm open for suggestions how this could be solved better.

### DIFF
--- a/net.go
+++ b/net.go
@@ -87,7 +87,14 @@ func incoming(c *client) {
 			break
 		}
 		DEBUG.Println(NET, "Received Message")
-		c.ibound <- cp
+		select {
+		case c.ibound <- cp:
+		case <-c.stop:
+			// This avoids a deadlock should a message arrive while shutting down.
+			// In that case the "reader" of c.ibound might already be gone
+			WARN.Println(NET, "incoming dropped a received message during shutdown")
+			break
+		}
 	}
 	// We received an error on read.
 	// If disconnect is in progress, swallow error and return


alsm:
I thought of something similar yesterday. I'm not too concerned about a packet that arrives as we've already initiated shutdown, the response to a qos 1 or 2 packet isn't sent until the packet is handled in `alllogic`, so would/should be resent at reconnect.
I am concerned though about packets arriving while we're in a blocking function in `alllogic` such as `persistInbound` and throwing that away.
I'm not going to merge this for now until I can do some more testing/thinking. Equally if you come up with any other ideas please contribute. I probably won't be able to do much on this until April due to going away.

databus23:
But this safeguard only kicks-in if ` <-c.stop` is returning something which means we are shutting down. If `c.ibound <- cp` blocks during normal operations that should be fine. As soon as `alllogic` enters a new loop it will read from `c.ibound` and the select in `incoming` will continue.

databus23:
I think the main problem is that as soon as `c.stop` is closed there is no guarantee that something will read from `c.ibound` anymore as `alllogic` might have already exited.

alsm:
Yes, of course, sorry I misread what you had done. In which case, as in my first comment, I'm not too concerned about potentially losing a message, qos 1/2 will be resent on reconnect, qos 0 has no delivery guarantee anyway (and often aren't queued up when clients are offline) so I don't think it'll have an impact on end users.
As such I am going to merge it, thanks for your work on this.

PR #87, opened 2017-03-27
Add an interface for reading config options from the client after initialization.

### LINKED ISSUE #5: Provide ability to retrieve options used to create client
migrated from Bugzilla [#486529](https://bugs.eclipse.org/bugs/show_bug.cgi?id=486529)
status UNCONFIRMED severity _normal_ in component _MQTT-Go_ for _1.2_
Reported in version _unspecified_ on platform _PC_
Assigned to: Al Stockdill-Mander

On 2016-01-26 03:08:32 -0500, Mark Mindenhall wrote:

> After creating a Client with ClientOptions, there's no way to retrieve the ClientOptions from the Client.  I want to implement a single ConnectionLostHandler to be used by all my connections (I have more than one), but when the handler is invoked, there's no way to tell from the *Client passed in which connection was lost.
> 
> Obviously, after a Client has been created, no further changes should be allowed to the ClientOptions, so the Client.options field should not be exported.  However, a method could be added that returns a _copy_ of the ClientOptions struct used to create the Client:
> 
> // Options returns a copy of the ClientOptions used to create this client.
> func (c *Client) Options() ClientOptions {
>   optionsCopy := c.options
>   return optionsCopy
> }

On 2016-01-27 02:35:29 -0500, Mark Mindenhall wrote:

> Actually, since the Servers field is an array of *url.URL, it's not so trivial as assigning to a new variable to copy the struct (I think the copy would retain the pointers).  Instead, the struct can be cloned via encoding/gob as follows (warning, may not be 100% correct, but gives general idea):
> 
> // Options returns a copy of the ClientOptions used to create this client.
> func (c *Client) Options() ClientOptions {
>   var buf = bytes.Buffer
>   enc := gob.NewEncoder(&buf)
>   enc.Encode(c.options)
> 
>   var optionsCopy *ClientOptions
>   dec := gob.NewDecoder(&buf)
>   dec.Decode(optionsCopy)
>   return *optionsCopy
> }

On 2016-01-27 10:57:21 -0500, Mark Mindenhall wrote:

> I think a better solution would be to create an interface comprised of getters for ClientOptions fields.  Then provide a method on the Client that returns the interface
> 
> Another solution here would be to create an interface comprised of getters for ClientOptions, then provide a method on Client that returns the interface.

On 2016-01-27 10:59:54 -0500, Mark Mindenhall wrote:

> I somehow submitted the previous comment before I was done editing...sorry for the redundancy, but you get the point.

aleksclark:
Should close #5 

Signed-off-by: Aleks Clark <aleks.clark@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -114,10 +114,11 @@ func NewClient(o *ClientOptions) Client {
 }
 
 func (c *client) AddRoute(topic string, callback MessageHandler) {
-    if callback != nil {
-        c.msgRouter.addRoute(topic, callback)
-    }
+	if callback != nil {
+		c.msgRouter.addRoute(topic, callback)
+	}
 }
+
 // IsConnected returns a bool signifying whether
 // the client is connected or not.
 func (c *client) IsConnected() bool {
@@ -587,6 +588,11 @@ func (c *client) Unsubscribe(topics ...string) Token {
 
 	DEBUG.Println(CLI, "exit Unsubscribe")
 	return token
+}
+
+func (c *client) OptionsReader() ClientOptionsReader {
+	r := ClientOptionsReader{options: &c.options}
+	return r
 }
 
 //DefaultConnectionLostHandler is a definition of a function that simply
--- a/options_reader.go
+++ b/options_reader.go
@@ -1,0 +1,138 @@
+/*
+ * Copyright (c) 2013 IBM Corp.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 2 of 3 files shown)


alsm:
Thanks for working on this, copying the contents of the struct before returning should stop anything inside the client from being edited while in use, except for the Store, Store is an interface, and for example the default memory store is passed into the ClientOptions as a pointer, so returning that would actually pass back the instance of the Store in use and allow someone to (stupidly I admit, you wouldnt ever want to) call the exported interface functions. I'm not sure if Store needs to be visible for clients once initialised?

aleksclark:
@alsm got it, thanks

apple314159:
See issue #108

PR #95, opened 2017-04-21
Update README: samples moved to cmd

else:
The sample moved to the `cmd` folder with 13afcbe8e41508479762a90e9242577210c2ca8d.


alsm:
Unfortunately ip validation must pass to be accepted, but thank you for noticing this, I have made the appropriate change

PR #98, opened 2017-04-27
If connection is lost before the corresponding ack packet

choonkeat:
Some `token.complete` chan would be left unclosed, blocking clients trying to `Wait()`

So when we encounter connection lost, we should iterate through messageIDs and close them all

### DIFF
--- a/messageids.go
+++ b/messageids.go
@@ -15,6 +15,7 @@
 package mqtt
 
 import (
+	"errors"
 	"sync"
 )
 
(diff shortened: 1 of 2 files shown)


chuyeow:
I ran into this same problem when the TCP connection is closed (unexpectedly) between sending `PUBLISH` packet and receiving the `PUBACK` packet. At this point, `token.Wait()` will block indefinitely (since it can never receive the `PUBACK` packet), causing my code to block indefinitely.

alsm:
It is not always true that when a connection is lost we want to close all the tokens, if you have a persistent session the server could resend, but otherwise you are correct that these should be cleaned up.

alsm:
Thanks for noticing this though, I have resolved the problem in the latest commit

PR #106, opened 2017-05-09
fix: Infinite '(digit & 128) == 1' will cause the dead loop

funnywwh:
If the RemainingLength is parsed, an infinite number of bytes of the highest order of 1 will result in a dead loop in the decodeLength function.

Signed-off-by: funnywwh <funnywwh@163.com>


alsm:
Thanks for this

PR #110, opened 2017-06-09
Update Token interface to make testing easier.

### LINKED ISSUE #107: Trouble Creating Mocks For Testing
I am trying to write unit tests for my project that is using this library.  For my unit tests I need to mock out the mqtt connection, so I tried to create a mock client object the implement the Client interface.  However many of the methods in the Client interface return a Token interface object.  The Token interface specifies a private method of flowComplete.  This makes it impossible to create a mock object that implements the Token interface, which makes it impossible to create a mock object that implements the Client interface.

Is there some other way to make this work for testing that I am missing?

Thank You

EricMaibach:
Moved the flowComplete method from the Token interface to the new
tokenCompletor interface.  The new tokenCompletor interface is
composed of the Token interface and the flowComplete method.  Having
the private flowComplete method in the public Token interface made
it impossible for applications using this library to create mock Token
objects for testing.  This change is intended to make testing easier
in applications that use this library.  This change is to resolve #107 

Signed-off-by: Eric Maibach <eric@maibachfamily.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -105,7 +105,7 @@ func NewClient(o *ClientOptions) Client {
 	}
 	c.persist = c.options.Store
 	c.status = disconnected
-	c.messageIds = messageIds{index: make(map[uint16]Token)}
+	c.messageIds = messageIds{index: make(map[uint16]tokenCompletor)}
 	c.msgRouter, c.stopRouter = newRouter()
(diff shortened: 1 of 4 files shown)


alsm:
Thanks for this, sorry it took so long to merge

PR #114, opened 2017-06-15
Add support for all_proxy env. var. for TCP and SSL/TLS/TCPS URL's

sandeepmistry:
On Linux, the `all_proxy` environment variable is used to share the SOCKS proxy information. (https://wiki.archlinux.org/index.php/proxy_settings#Using_a_SOCKS_proxy)

This PR makes the `tcp` and `tcps` use the SOCKS5 proxy if the environment variable is defined.

Note: `x/net/proxy` currently does not support passing in a timeout value for the dialer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,12 @@ This client is designed to work with the standard Go tools, so installation is a
 go get github.com/eclipse/paho.mqtt.golang
 ```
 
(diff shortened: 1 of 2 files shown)


alsm:
Thanks for this, really handy.

PR #139, opened 2017-08-27
Fix deadlock if callback invokes Subscribe

bonan:
In the `matchAndDispatch()` loop, a read lock is aquired when iterating through routes and matching.
Any callbacks found are currently invoked directly, with the read lock in place. If anyones tries to subscribe to additional topics (with callback) from a callback - there will be a deadlock since `addRoute` will try to aquire a write lock.

Fix: Call RUnlock() before invoking route callback to avoid deadlock if callback invokes Subscribe()

### DIFF
--- a/router.go
+++ b/router.go
@@ -135,10 +135,11 @@ func (r *router) matchAndDispatch(messages <-chan *packets.PublishPacket, order 
 			case message := <-messages:
 				sent := false
 				r.RLock()
+				handlers := []MessageHandler{}
 				for e := r.routes.Front(); e != nil; e = e.Next() {
 					if e.Value.(*route).match(message.TopicName) {
 						if order {
-							e.Value.(*route).callback(client, messageFromPublish(message))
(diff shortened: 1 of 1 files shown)


alsm:
Hi, thanks for the PR and spotting this issue, please sign the Eclipse CLA and sign off your submission so I can merge it, or let me know if this isn't possible and I'll fix the issue directly.

bonan:
I'm getting this message when trying to sign:
```Error, An error has occured and your Eclipse Contributor Agreement was not submitted. (ECA-03)```

I tried both yesterday and today without any luck.

I notified the license@ mailbox, who forwarded it to webmaster@ - no response from there yet

bonan:
The CLA issue is resolved, and I re-commited the fix with sign off

alsm:
Thanks for this.

daenney:
Any chance we could get a point release, 1.3.1 or something along those lines? It would be nice not to have to pin on master and resolving a deadlock would seem to warrant a release.

PR #144, opened 2017-09-22
Add an example to be more clear on what a broker URI should look like

tmcarr:
I just added a comment line with an example URI since it took me a bit to realise what the "unknown protocol" error from the net lib meant...


alsm:
👍 thanks

PR #148, opened 2017-10-30
Unix socket connection

gdunstone:
added functionality that allows users to open a connection to a unix socket.

- unix dialer as case option to openConnection switch

Signed-off-by: Gareth Dunstone <freedom.2.the.leetle.people@gmail.com>


alsm:
Cool, thanks.

PR #160, opened 2017-11-22
Fixes #158 changes KeepAlive to int64

### LINKED ISSUE #158: Change of KeepAlive to int64 breaks examples
https://github.com/eclipse/paho.mqtt.golang/blob/512dcaa501c1580ea8eccb062ba06cc0c2b78172/options.go#L55

```
thughes@titanium [0] $ go get github.com/eclipse/paho.mqtt.golang/cmd/...
# github.com/eclipse/paho.mqtt.golang/cmd/stdoutsub
eclipse/paho.mqtt.golang/cmd/stdoutsub/main.go:65:12: cannot use 30 * time.Second (type time.Duration) as type int64 in field value
```

timhughes:


alsm:
Thanks, I fixed this by changing the code to use NewClientOptions() instead as by just crafting a struct I was missing required options

PR #161, opened 2017-12-01
Add timeout for publish messages. Drop message if timeout expired to prevent memory leaks.

platinumthinker:
Hi! I have a problem when i work with buggy mqtt server: sometimes server allow to establish tcp connection but does not send Ack on Connect message.
I have many goroutines for publishing message in go client.
This commit fixes it.


alsm:
I see the issue, but this is probably a feature that should be disableable

PR #171, opened 2017-12-25
Fix panic on disconnect

antlad:
Add check on stop route chan close


rollulus:
It would be great if this gets merged actually. Ran into the same issue.

Issue #16, opened 2016-02-09
The acknowledgement for incoming messages are sent even if the associated MessageHandler panics


jpwsutton:
migrated from Bugzilla [#469350](https://bugs.eclipse.org/bugs/show_bug.cgi?id=469350)
status UNCONFIRMED severity _major_ in component _MQTT-Go_ for _1.2_
Reported in version _0.9_ on platform _Macintosh_
Assigned to: Al Stockdill-Mander

On 2015-06-04 02:50:04 -0400, Saikrishnan Ranganathan wrote:

> It looks like the acknowledgement for inbound QoS 1 and 2 messages are sent to the broker even if the MessageHandler associated with the topic panics. So when the client is next re-connected, any QoS 1 or 2 messages will NOT be redelivered by the broker.

gleb-chipiga:
+1

PR #175, opened 2018-01-04
Fix: The acknowledgement for incoming messages are sent even if the associated MessageHandler panics #16

xeizmendi:
This is a proposed fix for #16

Unfortunately I couldn't think of sensible way of solving the problem without breaking the API and allowing MessageHandler to return an error.

Given that the network and message routing layer are clearly separate with this implementation the ACK execution is delegated by the network to the message router so it is only called if all the handlers process the message successfully.


dambrisco:
@alsm Any movement/comments on this PR?

PR #192, opened 2018-03-12
Add a way to set the username and password via callback.

vaelen:
This is useful for Google IoT Core and other systems which use authentication tokens that expire.

### DIFF
--- a/message.go
+++ b/message.go
@@ -88,13 +88,19 @@ func newConnectMsgFromOptions(options *ClientOptions) *packets.ConnectPacket {
 		m.WillMessage = options.WillPayload
 	}
 
-	if options.Username != "" {
+	username := options.Username
+	password := options.Password
+	if options.CredentialsProvider != nil {
+		username, password = options.CredentialsProvider()
+	}
+
+	if username != "" {
 		m.UsernameFlag = true
-		m.Username = options.Username
+		m.Username = username
 		//mustn't have password without user as well
-		if options.Password != "" {
+		if password != "" {
 			m.PasswordFlag = true
-			m.Password = []byte(options.Password)
+			m.Password = []byte(password)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


danielbprice:
This would solve a lot of issues for me, I think.

vaelen:
I updated the code to provide both the username and password, since that seems more useful in the general case. 

I'll add a test case in my next commit.  There doesn't appear to be a test for this particular bit of code yet.

vaelen:
Test case added.

alsm:
Great solution, thanks for this

danielbprice:
This is awesome!  I'll be changing my code to use it this week.

vaelen:
Glad you liked it. :smile: Here is the code where I'm using it, for reference: https://github.com/vaelen/iot

danielbprice:
Neat!  I will drop you a note in an issue over there.  It looks really nice.  I think you should add it to "awesome-go" or something, as I really had hoped to have something like this three months ago.

ovaltzer:
Quick question: while using jwt token for mqtt, with Google iot core, the broker automatically disconnect the cline when the token expires. How would you recommend recreating the token to be used in the connection before expiring? Is there any way of doing it currently?

vaelen:
You have to reconnect, but you can let Paho's auto-reconnect handle it.  On reconnect you need to resubscribe to the config topic though.  My Google IoT client library does this, check out https://github.com/vaelen/iot/blob/master/thing.go#L72 for more details.

ovaltzer:
I used the auto reconnect for generating a jwt, but it's just makes we wonder if this is the correct behavior while there is no real reason to reconnect but instead just dynamically update the jwt token which the client uses

vaelen:
Yeah, I agree that it would make sense to update the token automatically.

PR #194, opened 2018-03-15
Implements store resume

### LINKED ISSUE #90: Read FileStore after application restart
The documentation of the FileStore says 

> FileStore implements the store interface using the filesystem to provide true persistence, even across client failure

but I could not find how to send the stored messages after the application restarts, could you please tell me how is it done?

ghost:
Actually read stored message when connecting or reconnecting and send them. Fixes #90

### DIFF
--- a/client.go
+++ b/client.go
@@ -12,6 +12,8 @@
  *    Mike Robertson
  */
 
+// Portions copyright © 2018 TIBCO Software Inc.
+
 // Package mqtt provides an MQTT v3.1.1 client library.
 package mqtt
 
@@ -55,6 +57,9 @@ type Client interface {
 	// IsConnected returns a bool signifying whether
 	// the client is connected or not.
 	IsConnected() bool
+	// IsConnectionOpen return a bool signifying wether the client has an active
+	// connection to mqtt broker, i.e not in disconnected or reconnect mode
+	IsConnectionOpen() bool
 	// Connect will create a connection to the message broker, by default
 	// it will attempt to connect at v3.1.1 and auto retry at v3.1 if that
 	// fails
@@ -123,6 +128,8 @@ func NewClient(o *ClientOptions) Client {
 	switch c.options.ProtocolVersion {
 	case 3, 4:
 		c.options.protocolVersionExplicit = true
+	case 0x83, 0x84:
+		c.options.protocolVersionExplicit = true
 	default:
 		c.options.ProtocolVersion = 4
 		c.options.protocolVersionExplicit = false
@@ -163,6 +170,21 @@ func (c *client) IsConnected() bool {
 	}
 }
 
+// IsConnectionOpen return a bool signifying whether the client has an active
+// connection to mqtt broker, i.e not in disconnected or reconnect mode
+func (c *client) IsConnectionOpen() bool {
+	c.RLock()
+	defer c.RUnlock()
+	status := atomic.LoadUint32(&c.status)
+	switch {
+	case status == connected:
+		return true
+	default:
+		return false
+	}
+}
+
+
 func (c *client) connectionStatus() uint32 {
 	c.RLock()
 	defer c.RUnlock()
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -599,6 +604,67 @@ func (c *client) SubscribeMultiple(filters map[string]byte, callback MessageHand
...
+	for idx, key := range storedKeys {
+		packet := c.persist.Get(key)
+		details := packet.Details()
+		if isKeyOutbound(key) {
+			switch packet.(type) {
+			case *packets.SubscribePacket:
+				if subscription {
+					DEBUG.Println(STR, fmt.Sprintf("loaded pending subscribe (%d)", details.MessageID))
+					token := newToken(packets.Subscribe).(*SubscribeToken)
+					tokens[idx] = token
```
danielbprice: I'm unclear why the list of tokens returned is not a slice with no nils?  That is to say, why not simply append tokens as needed?  The code as written seems to produce some Tokens which are nil.

options.go
```diff
@@ -118,6 +120,12 @@ func (o *ClientOptions) AddBroker(server string) *ClientOptions {
 	return o
 }
 
+// SetResumeSubs will enable resuming of stored (un)subscribe message
```
danielbprice: A comment which mentions the relationship to CleanSession would be helpful.

client.go
```diff
@@ -387,6 +386,8 @@ func (c *client) reconnect() {
 	go alllogic(c)
 	go outgoing(c)
 	go incoming(c)
+
+	c.resume(false)
```
danielbprice: Maybe I have misunderstood the code, but it seems weird that resume() returns a set of tokens, but there's no opportunity to wait on them.  I don't know how that might best be fixed.  Maybe the tokens could be stored in the client object and available for retrieval via some function call?

client.go
Illule: Indeed this array may contains nil token. At the time I choose this to avoid memory allocation behind the append function (if I understood correctly how the append function works). I think it costs more allocating memory than checking if an element is nil. But I agree it will be more convenient if there was no nil. I can change it if convenience is prefered to performance.

client.go
Illule: I confess I wrote this patch quickly. I needed this for a project with a very short deadline. And as it is, the resume function should not return anything as the array of token is never used. Moreover I was not sure of the wanted behaviour for this library. Should I write a function to wait for all token being completed ? Let the user access this array of token and let him deal with it ? Call a user defined callback when all tokens are completed ? Or removes this array of token ?

I hope my explanations are clear. English is not my native language

danielbprice: I think those are great questions.  If I have a chance I will go survey the other MQTT libraries and see what their strategy is.

client.go
danielbprice: It should be an extremely rare operation, with a lot of cost (such as pulling things off of disk) anyway.  So I would prioritize a clean interface over saving a few cycles.

Illule: In fact in my use case it is not rare. I am using this on a mobile device which is not always connected. This feature allows me to backlog all published message on the go once returned near the station where it reconnects to the broker. As my device is battery powered I tend to minimize usage of cpu. But I guess my usage is quiet specific and not suited for a generic library. I will make the changes depending of the result of the discussion about what to do with this array.

client.go
alsm: The idea behind the tokens was to allow you do effectively do synchronous operations with an async messaging system, so you can do a send of a QoS2 message and block until the message is full sent and acknowledged by the server. If we've been interrupted and are reloading this data from the Store the original tokens have already been closed and there is no easy way to associate new tokens with the original actions. If you can't think of a good reason or way to pass this back to the user I'm happy for now to just swallow them inside the library.

Illule: Ok, juste let me removes the tokens array, if it is not used there is no need to allocate a eventually 2^16 length array.

Issue #197, opened 2018-03-22
Panic on Disconnect


amenzhinsky:
Hi there!

When a broker address is not valid the following code panics, but it shouldn't:
```
package main

import (
	"log"
	"os"

	"github.com/eclipse/paho.mqtt.golang"
)

func main() {
	mqtt.ERROR = log.New(os.Stdout, "[ERROR] ", 0)
	mqtt.WARN = log.New(os.Stdout, "[WARN]  ", 0)
	mqtt.DEBUG = log.New(os.Stdout, "[DEBUG] ", 0)

	o := mqtt.NewClientOptions()
	o.AddBroker(":1882") // invalid url
	c := mqtt.NewClient(o)
	if t := c.Connect(); t.Wait() && t.Error() != nil {
		panic(t.Error())
	}
	c.Disconnect(250)
}
```

Output:
```
[DEBUG] [client]   Connect()
[DEBUG] [store]    memorystore initialized
[ERROR] [client]   Failed to connect to a broker
[DEBUG] [store]    memorystore closed
[WARN]  [client]   Disconnect() called but not connected (disconnected/reconnecting)
panic: close of nil channel

goroutine 1 [running]:
github.com/eclipse/paho%2emqtt%2egolang.(*client).closeStop(0xc4200f8a80)
	github.com/eclipse/paho.mqtt.golang/client.go:482 +0x12f
github.com/eclipse/paho%2emqtt%2egolang.(*client).disconnect(0xc4200f8a80)
	github.com/eclipse/paho.mqtt.golang/client.go:495 +0x2f
github.com/eclipse/paho%2emqtt%2egolang.(*client).Disconnect(0xc4200f8a80, 0xfa)
	github.com/eclipse/paho.mqtt.golang/client.go:439 +0x1ce
main.main()
	main.go:14 +0xb9
```

It happens because `url.Parse` silences an error [here](https://github.com/eclipse/paho.mqtt.golang/blob/36d01c2b4cbeb3d2a12063e4880ce30800af9560/options.go#L120)  and this can be really confusing, since `:1883` looks like a valid address.

It's not very clear to me what we should do in this situation: 
1. `AddBroker` panics on error.
1. `Connect` returns the `No brokers available` error.
1. Log the error with `ERROR.Printf("Unable to parse broker address: %s", err)`.

Also we can use `net.ResolveTCPAddr` as a fallback that can be converted to an url like:  `tcp://address:port`

[...]

PR #199, opened 2018-03-23
Improve broker address parsing

amenzhinsky:
See #197

### DIFF
--- a/options.go
+++ b/options.go
@@ -17,6 +17,7 @@ package mqtt
 import (
 	"crypto/tls"
 	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


alsm:
The code you've removed from client.go was used to ensure that in the situation where we have multiple servers to attempt to connect to we always revert to trying to connect at version 4 first if there was no explicit protocol version set.
If we attempt a connection at version 4, fail, retry at version 3, fail and go to the next server, we'll be stuck at version 3 only.

amenzhinsky:
@alsm sorry for that, didn't make any sense to me at the first glance, thought it was some kind of legacy code. Reverted it.

alsm:
👍

PR #207, opened 2018-04-09
v5 alpha

alsm:


paho/client.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// Connect is used to connect the client to a server. It presumes that
+// the Client instance already has a working network connection.
+// The function takes a pre-prepared Connect packet, and uses that to
+// establish an MQTT connection. Assuming the connection completes
+// successfully the rest of the client is initiated and the Connack
+// returned. Otherwise the failure Connack (if there is one) is returned
+// along with an error indicating the reason for the failure to connect.
+func (c *Client) Connect(cp *p.Connect) (*p.Connack, error) {
```
dolanor: Why not pass directly the connect packet options directly to the Connect() func. Has the connect packet another use than to be passed to Connect()? Otherwise, I feel it's an implementation detail that burdens the API.

Same comment for Disconnect(), Publish(), Subscribe()

alsm: For flexibility, the option funcs to configure the packets don't currently cover everything you might want to do, and as the fields of the packets are all exported from the `packets` library you are just as free to edit them directly when constructing a packet.

paho/client.go
```diff
@@ -0,0 +1,378 @@
...
+
+	return ca.Content.(*p.Connack), nil
+}
+
+// Incoming is the Client function that reads and handles incoming
+// packets from the server. The function is started as a goroutine
+// from Connect(), it exits when it receives a server initiated
+// Disconnect, the Stop channel is closed or there is an error reading
+// a packet from the network connection
+func (c *Client) Incoming() {
```
dolanor: Why put it as a public function?
It seems it is just an internal implementation detail, and calling it twice might create havoc.

paho/client.go
dolanor: And why not directly parameters to the function?
What about zero values with sensible default ?

paho/client.go
```diff
@@ -0,0 +1,378 @@
...
+func (c *Client) Connect(cp *p.Connect) (*p.Connack, error) {
+	debug.Println("Connecting")
+	c.Lock()
+	defer c.Unlock()
+
+	if c.Conn == nil {
+		return nil, fmt.Errorf("Client has no connection")
+	}
+
+	cp.ProtocolName = "MQTT"
```
dolanor: What if the packet is `nil`?

paho/client_options.go
```diff
@@ -0,0 +1,148 @@
...
+		c.MIDs.Clear()
+		return nil
+	}
+}
+
+// WithPersistence sets the client's internal message persistence mechanism
+// to be the Persistence passed into this function. If this option is not used
+// *no* persistence mechanism is used. This library currently also provides a
+// memory based Persistence.
+func WithPersistence(p Persistence) func(*Client) error {
```
amenzhinsky: This new version is minimalistic and I really like it, do you think that it should implement message persistence? As I see it'll be needed to write a higher-level abstraction per each project for features like: reconnecting, lazy connection, messages buffering, etc.

alsm: Thanks, the persistence mechanisms are pluggable, I haven't looked to provide a proper one yet as they can be a bit of a pain to debug, although if you were interested in doing one, either as a PR to the main code or a standalone library that others can use, that'd be cool

alsm: Some features like autoreconnect, message buffering etc from the current v3 client don't seem to actually make it any simpler for the application writer, and add more code for me to maintain. Happy to look at any such features and see if their is value in them and how they might be implemented in the v5 client

alsm:
The latest update just now is an alpha2, this removes the need to import the packet library into your app and provides compile checking of properties in packets you send to the server (previously I just dropped properties that were set and not applicable to the packet being sent).

Any more comments, suggestions or PRs welcome.

My intention is to freeze the current v3 client as far as API goes and only accept bugs/bugfix changes. Ultimately once the v5 API looks to be done I would want to do a new v3 client based on the same design.

paho/client_options.go
kdsch: @alsm Forgive me if I am not understanding the reason for this design; nevertheless I see room for improvement.

Function values are not comparable. As a consequence, functions that return functions cannot be tested in a data-driven way (there's no declarative way to ensure the output is correct; you have to call the function and test its behavior). I'm seeing them liberally used here, and this style of code is not typical in Go. Granted, the functions seem straightforward. Their role is to configure the client. But why not let that be a data declaration? Why does it have to involve higher-order functions? The fields being initialized are public anyway.

A Go motto is to accept interfaces and return structs. A function is more like an interface than a struct; it can only be called.

alsm:
Please see https://github.com/eclipse/paho.mqtt.golang/pull/237 for more current v5 client development

PR #214, opened 2018-05-10
Fix: The acknowledgement for incoming messages are sent even if the associated MessageHandler panics #16

xeizmendi:
This is a proposed fix for #16

In this fix we delegate sending acks back to the user of the library by adding an Ack function to the Message interface. This means acks are only sent whenever we are really done processing the message.

### DIFF
--- a/message.go
+++ b/message.go
@@ -16,6 +16,7 @@ package mqtt
 
 import (
 	"github.com/eclipse/paho.mqtt.golang/packets"
+	"sync"
 )
 
 // Message defines the externals that a message implementation must support
@@ -28,6 +29,7 @@ type Message interface {
 	Topic() string
 	MessageID() uint16
 	Payload() []byte
+	Ack()
 }
 
 type message struct {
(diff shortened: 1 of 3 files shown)


xeizmendi:
It's been almost a month since this PR was created.
Will it be reviewed at some point?

aminehmida:
Any news regarding this PR?

gm42:
Is there some documentation about how to use `Ack()`? I could not find any, nor any example.

I assume also that there is no `Nack()` equivalent.

amuldowney:
I don't understand how this change fixes issue #16 . It looks like matchAndDispatch calls .Ack() after the handler is called no matter what. It doesn't seem like there is a way to fail a message.

xeizmendi:
@gm42 @amuldowney unfortunately there is no way to fail a message, the MQTT protocol doesn't support this, it will eventually be sent again when the message times out. I stopped using MQTT because of this.

@aminehmida about the call to Ack, this was added later on by someone else because my PR changed the default behaviour. As in, you always need to acknowledge or else you will receive the message over and over again, you'll notice that in this PR Ack is not always called.
The correct way, I guess, would be to add an AutoAck option to the client so it's configurable and make it the default option.

aminehmida:
@xeizmendi thanks for the clarification.

PR #216, opened 2018-05-15
add IsConnectionOpen method to client interface

ovaltzer:
This PR adds a method to the client interface to which helps the client consumer to understand if the current connection is open (vs IsConnected which returns true in case of reconnect state).
this helps consumers to decide if they want to try and publish events which they will know that there is low chance that will be sent, and also not make the outbound channel full (and therefore blocking).
it helps for iot devices which might be disconnected from internet for some time and come back later...


odedva:
please close this pr opening a new one with valid eclipse account

PR #217, opened 2018-05-18
Add is connection open method to interface

odedva:
This PR adds a method to the client interface to which helps the client consumer to understand if the current connection is open (vs IsConnected which returns true in case of reconnect and even connecting state - which is quite misleading IMHO).
this helps consumers to decide if they want to try and publish events which they will know that there is low chance that will be sent, and also not make the outbound channel full (and therefore blocking).
it helps for iot devices which might be disconnected from internet for some time and come back later...

### DIFF
--- a/client.go
+++ b/client.go
@@ -55,6 +55,9 @@ type Client interface {
 	// IsConnected returns a bool signifying whether
 	// the client is connected or not.
 	IsConnected() bool
+	// IsConnectionOpen return a bool signifying wether the client has an active
+	// connection to mqtt broker, i.e not in disconnected or reconnect mode
+	IsConnectionOpen() bool
 	// Connect will create a connection to the message broker, by default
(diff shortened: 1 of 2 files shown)


alsm:
This behaviour only applies if the autoreconnect function is enabled, the idea being that we hide transitory networking issues. Tbh it might not have been the best idea but if you don't like it you can disable autoreconnect and handle reconnecting when the connection is lost in the ConnectionLostHandler

ovaltzer:
Actually the reconnect is perfect for me as i don't need to bother on that with my app.
the thing is, I prefer to avoid being "hand" with too many messages in the client channels (as i saw examples that the client hangs when the channels are full and  there is no connectivity and i would like to avoid it.  therefore i store the messages locally in the app in case that the client is in other state then connected, and only try to publish when connected...

Issue #203, opened 2018-03-29
Can't implement Token interface.


vaelen:
I am trying to create unit tests which use a mock implementation of the mqtt.Client interface, but in order to do so I must also create a mock implementation of the mqtt.Token interface.  However, the mqtt.Token interface has an unexported method: `setError(error)`, which means that only types in the `mqtt` package can implement the interface.

My question is: would anyone object to me changing `setError(error)` to `SetError(error)`?

estutzenberger:
I also just ran into this issue. My solution was to leave `setError` private and put it in it's own interface. I can see reasons to hide the `setError` function but... it makes it impossible to mock.
However, that function seems to be needed only by `tokenCompleter`. Locally the change I made looks like:

``` go
//Token defines the interface for the tokens used to indicate when
//actions have completed.
type Token interface {
    Wait() bool
    WaitTimeout(time.Duration) bool
    Error() error
}

type TokenErrorSetter interface {
    setError(error)
}

type tokenCompletor interface {
    Token
    TokenErrorSetter
    flowComplete()
}
```

alsm:
Closing as linked PR was merged

PR #221, opened 2018-05-24
Slight adjustment to make Token interface mockable

estutzenberger:
Signed-off-by: Eric Stutzenberger <eric.stutzenberger@rigado.com>

See #203 for details. This may not be the best solution, but it has very low code impact as it requires no change to the main source.


aphistic:
I'd love to see this get merged. The unexported interface method is also preventing me from mocking the interface.

PR #224, opened 2018-06-20
Added Bridge Protocol Version indicator

linux4life798:
This commit adds an unofficial ProtocolVersion flag that indicates to the Broker that you are serving as a bridge.
The Broker then does not echo messages published by the bridge, back to the bridge.
Please see https://github.com/mqtt/mqtt.github.io/wiki/bridge_protocol .

This is confirmed to be implemented and working using Mosquitto.


linux4life798:
My mistake. This feature seems to already exist in the newer version.

PR #237, opened 2018-09-04
V5

alsm:


amenzhinsky:
Just a general question, why wouldn't we follow the approach proposed by the go team to create sub-folders for major versions. https://blog.golang.org/versioning-proposal.

alsm:
Given the very large differences between the current code and the new client, I may just go for a whole new package name and repo.

alsm:
Development of the v5 client (and future v3 in the same API style) is now in this repo https://github.com/eclipse/paho.golang please help develop it.

Issue #226, opened 2018-06-28
Panic from package files net.go and client.go


Strosel:
When I receive, or send a message from/to my server I (sometimes, even though the messages are identical) get  the following panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1e68cc]

goroutine 228106 [running]:
github.com/eclipse/paho%2emqtt%2egolang.outgoing(0x11eee000)
	/Users/axpa/go/src/github.com/eclipse/paho.mqtt.golang/net.go:204 +0xb54
created by github.com/eclipse/paho%2emqtt%2egolang.(*client).reconnect
	/Users/axpa/go/src/github.com/eclipse/paho.mqtt.golang/client.go:412 +0x1124
```

as you can see it doesn't trace back to my own project and would therefore seem to be a bug within the package itself

ricardoatsouza:
The same started happening to me as well. I noticed it would happen when the incoming and outgoing of data is considerable big. In my case, I am speaking of a service that receives, on average, ~ 1k records per second (~ 1Mb/s) and produces ~ 20 k records per second (~ 10 MB/s).

This service in particular is subscribed to 1 topic (only one subscription), and has 10 producer that will send data to the same output topic.

First question that comes to my mind is: would this issue be related to a race condition?
Second is: when using multiple connections in the same service, are they sharing any resources?

alsm:
Closing as PR merged

PR #239, opened 2018-09-18
Adding a nil check to net.go.

ricardoatsouza:
Quick and easy fix for when an error happens while writing to [`msg.p`](https://github.com/eclipse/paho.mqtt.golang/compare/master...ricardoatsouza:bugfix/issue-%23226_panic_from_net.go?expand=1#diff-e2893928b1f232947cf45a36fdcb6f9bR203)

This has been reported as an issue already: https://github.com/eclipse/paho.mqtt.golang/issues/226

A `token` is mostly created in the file `client.go` and, when created from there, is never `nil`. However, it is also created by the `net.go` script (lines [268](https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L268), [280](https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L280), [302](https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L302) and [311](https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L311)), and in all these cases, token (variable `t`) is `nil`.

As a consequence, if an error happens, there is no `t` to call `setError`, and it panics.

### DIFF
--- a/net.go
+++ b/net.go
@@ -201,7 +201,9 @@ func outgoing(c *client) {
 			DEBUG.Println(NET, "obound priority msg to write, type", reflect.TypeOf(msg.p))
 			if err := msg.p.Write(c.conn); err != nil {
 				ERROR.Println(NET, "outgoing stopped with error", err)
-				msg.t.setError(err)
+				if msg.t != nil {
+					msg.t.setError(err)
+				}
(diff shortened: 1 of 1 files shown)


alsm:
Thanks for this :)

PR #249, opened 2018-10-30
Fixed "close of nil channel" when calling Stop() on a not connected client

an2deg:
Hi!

I have a panic "close of nil channel" when I was calling Stop() on a not connected Client.
There is a simple fix for this.


alsm:
👍

PR #252, opened 2018-11-01
Check string length before slicing file names in filestore.

FabianInostroza:


FabianInostroza:
If you or other software creates a file with short name in the file store folder then the program crashes when loading the file store

PR #253, opened 2018-11-02
Check file name length before slicing.

FabianInostroza:
If there is file with a short name the software would crash when attempting to slice the file name.

Signed-off-by: Fabián Inostroza <fabianinostroza@udec.cl>

### DIFF
--- a/filestore.go
+++ b/filestore.go
@@ -166,6 +166,10 @@ func (store *FileStore) all() []string {
 	for _, f := range files {
 		DEBUG.Println(STR, "file in All():", f.Name())
 		name := f.Name()
+		if len(name) <= 4 {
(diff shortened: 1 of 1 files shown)


dambrisco:
@FabianInostroza I believe the correct check here would be `< 4`, not `<= 4`. A string of length 4 can be sliced by the following block just fine.

MattBrittan:
This PR has been silent for quite a while (due to the lack of response to comment from @dambrisco). I agree that this is an issue but feel a better solution would be to avoid using a constant (4) because that assumes that `msgExt` is fixed. I'd prefer something along the lines of:

`if len(name) < len(msgExt) || name[len(name)-len(msgExt):] != msgExt`

MattBrittan:
Implemented my version of this in PR523.

PR #257, opened 2018-11-14
fix nil pointer by injecting list element pointer into goroutine scope in matchAndDisptch

rmn87:


dambrisco:
Looks like this might be resolved by 103f6a1 using a different method.

alsm:
Thanks for the PR, sorry I ddin't notice it before I fixed it myself

PR #270, opened 2018-12-17
Handle errors when reading packets

akokhanovskyi:
The current implementation interprets the packet content as it reads it from the io.Reader (typically backed by net.Conn). At the same time, almost no error handling is present. If you're lucky, everything works fine. But in more challenging scenarios (like a heavily loaded server) network mishaps, disconnects, and timeouts lead to the code trying to interpret data segments that in fact were not read from the network. This, in turn, leads to spooky errors and laborious investigations. For example, a Publish read timeout is often reported as "Error unpacking publish, payload length < 0" - caused by not initialized `RemainingLength` in the `FixedHeader`.

Likely, more work is required to handle all error conditions, but I would like to receive some community feedback on an attempt to handle/report errors where they actually occur.


alsm:
thanks for doing all that

PR #275, opened 2019-01-14
Shared subscription 

maheshjena1:
The shared subscription feature for the subscription topic with keyword "$share/" is not working as expected.


alsm:
Sorry, you're trying to merge in code from the v5 branch into main, these are not designed to mix. Is there a PR you meant to make?

PR #283, opened 2019-02-24
Add callback prior to reconnection attempt

ajhattozxcorp:
Added an optional 'OnReconnecting' callback in the client options, which can be used to modify the client options just prior to reconnection attempts.

In our particular case, we needed to do this to update custom authentication tokens in the HTTP headers section, as our authentication tokens can expire and need renewing before reconnection attempts.

### DIFF
--- a/client.go
+++ b/client.go
@@ -345,8 +345,8 @@ func (c *client) reconnect() {
 	)
 
 	for rc != 0 && c.status != disconnected {
-		if nil != c.options.OnWillReconnect {
-			c.options.OnWillReconnect(c, WillReconnectArgs{&c.options})
+		if nil != c.options.OnReconnecting {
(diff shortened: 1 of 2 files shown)


alsm:
I had started doing something more specific to just HTTP Headers, but this would be fine too. I think it should just be named reconnecthandler as the will prefix could be confused to be related to the mqtt will facility. Also I'm not sure the ClientOptions passed to the handler func need wrapping in another struct. You will also need to register with eclipse and sign the CLA to pass the check. Thanks

ajhattozxcorp:
@alsm I've made some updates inline with your suggestions and signed the Eclipse Contributor Agreement.

alsm:
Thanks for the update on this, apparently you need to sign off your commits for it to pass the eca check.

PR #290, opened 2019-03-22
Resending qos packages after reconnecting

### LINKED ISSUE #278: Resending of QoS 1 after losing connection doesn't work
I'm using AutoReconnect=true, FileStore and CleanSession=false.

1) If remove my internet connection
2) and send a QoS 1 message
3) and then resume the internet connection the MQTT client reconnects as expected, however the messages are lost.

However, if I kill my client after step 2 and after I have resumed the internet connection start the client it will resend my QoS 1 messages.

I would assume that QoS 1 messages should be resent in both cases.

cuckoohello:
resolves #278

Code was accidentally deleted by commit 66c89a7

Signed-off-by: Yafeng Shan <syafeng@infervision.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -429,6 +429,8 @@ func (c *client) reconnect() {
 	go alllogic(c)
 	go outgoing(c)
 	go incoming(c)
+
+	c.resume(false)
 }
 
 // This function is only used for receiving a connack


alsm:
thanks

lglenat:
Shouldn't it be done only if `cleanSession` is not `true`?

```
// Take care of any messages in the store
if c.options.CleanSession == false {
    c.resume(c.options.ResumeSubs)
} else {
    c.persist.Reset()
}
```

cuckoohello:
@lglenat not needed. Because it's reconnect not a new connect.

Since it's reconnect, it means the same connection, so we can just resume the qos packages.

So  I just commit back the code accidentally deleted by commit 66c89a7.

lglenat:
@cuckoohello ok this makes sense. Thanks for the clarification.
Also there might be an issue with the `resume()` function see https://github.com/eclipse/paho.mqtt.golang/issues/291 that I just created

dambrisco:
@cuckoohello @alsm Is there any particular reason this is resumed with `false` instead of `c.options.ResumeSubs`? I'm working on a change that touches this and the [original commit](https://github.com/eclipse/paho.mqtt.golang/commit/d651ccc84e5f8fbe2811d5671bd0c15b94a4d4c0) doesn't clarify why this is done, so I'm just looking for context.

PR #293, opened 2019-03-26
Handle failure ReturnCode in SubackPacket

sko00o:
ReturnCodes not only contain qoss. see [MQTT Version 3.1.1 - SUBACK](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718068)
`ReturnCode == 0x80` indicates subscribe failure. If mqtt server enable ACL plugins, someone subscribe write-only topics should get error in ` token.Error()`.

### DIFF
--- a/net.go
+++ b/net.go
@@ -252,6 +252,13 @@ func alllogic(c *client) {
 				case *SubscribeToken:
 					DEBUG.Println(NET, "granted qoss", m.ReturnCodes)
 					for i, qos := range m.ReturnCodes {
+						// ReturnCodes not only contain qoss,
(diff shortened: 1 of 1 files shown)


alsm:
Unfortunately as you can subscribe to multiple topics at the same time, it wouldn't be strictly correct to return an error if only one of the subscriptions failed. I see your point but do not intend to change this behaviour, the 0x80 code is passed back to the caller and as the response codes must be in the same order as the topics in the subscribe it is possible to work out which one failed.

PR #299, opened 2019-04-02
Don't clean up if AutoReconnect is true

### LINKED ISSUE #296: Publish handling during disconnected state differs
When attempting to Publish a message in a disconnected state, there are two distinct cases:

A. The broker is offline but can be located.
B. The broker is unreachable.

Case A results in the Publish attempt idling and, under normal circumstances, successfully publishes once the broker is back online. Calling Wait() on the PublishToken blocks as expected until this completes and Error() returns a nil value. The client internally reports an EOF but does not pass this along with the token.

Case B results in the Publish attempt idling as well, but unless a connection is reestablished within ~45 seconds the attempt fails (the ~45 second window does NOT appear to be controlled by any of the timeout values that can be set via ClientOptions). Calling wait on the PublishToken blocks, but Error() returns with `"Connection lost before Publish completed"` is returned. The client internally reports `"pingresp not received, disconnecting"`, which leads to the aforementioned Error() value. Subsequent publish attempts block in a similar manner to Case A.

In both cases, the client reports `"error triggered, stopping"` and `"logic stopped"` following the respective EOF or pingresp messages. The key difference appears to be that prior to the Publish call, the connection state for Case B is "connected".

Both cases should use Case A's behavior - the Publish attempt should idle and be reattempted once a connection is successfully reestablished.

This is something I'm currently trying to fix, so advice/guidance on accomplishing this would be greatly appreciated (or a persuasive argument that Case B is correct and expected behavior). It seems that this must be strictly related to `internalConnLost` function where the connection status is checked, and the necessary change would be to treat a Publish that fails due to a connection loss as though it were called in the `reconnecting` state if AutoReconnect is set to true.

dambrisco:
Fixes #296 - **possibly** out of spec. A fully in-spec fix may require that **only** PUB/SUB/UNSUB messages that had **never** reached the server (e.g. encountered any issue other than a straight dial/networking timeout/server unreachable) be continued following a reconnect.

However, `AutoReconnect` is not technically spec anyway, and because these continuations **only** affect clients using `AutoReconnect` I personally feel this is an acceptable and logical fulfillment of the AutoReconnect feature.

To accomplish this fully in-spec, I would recommend providing additional information about the flow's state via the `Token` for `cleanUp` to analyze, but that would require far more significant code changes than this fix  and introduce a significant amount of cyclomatic complexity to what's currently a clean and simple function - and given what I've mentioned above about the `AutoReconnect` flag, I'm unsure that would be necessary or even wise.

### DIFF
--- a/client.go
+++ b/client.go
@@ -503,7 +503,7 @@ func (c *client) internalConnLost(err error) {
 		c.closeStop()
 		c.conn.Close()
 		c.workers.Wait()
-		if c.options.CleanSession {
+		if c.options.CleanSession && !c.options.AutoReconnect {
 			c.messageIds.cleanUp()
 		}
 		if c.options.AutoReconnect {


alsm:
Thanks for catching this.

PR #301, opened 2019-04-11
Add ForcePersist option and ForceResume function

dambrisco:
Adds the ability to prevent clearing messages from a persistent store when calling `Connect` which can then be resumed using `ForceResume`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -91,6 +91,9 @@ type Client interface {
 	// OptionsReader returns a ClientOptionsReader which is a copy of the clientoptions
 	// in use by the client.
 	OptionsReader() ClientOptionsReader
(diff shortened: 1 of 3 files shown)


alsm:
Given that there are no guarantees about what would happen in some situations with this providing functionality with indeterminate results I don't think is a good idea.

dambrisco:
@alsm Sorry to necro this PR but I don't think I noticed when it was originally closed. Could you elaborate on the indeterminate results you're concerned about? I believe you're indicating that the ForceResume function cannot be called safely in some states, with which I'd agree. It's something I'm happy to address if this functionality is otherwise acceptable.

PR #302, opened 2019-04-13
Error "Error at readfull" when disconnect 

UselezzProgrammer:
Fix error "Error at readfull"

Now when I just try to connect to a server and then disconnect, I will get an "Error at readfull" error. It is not graceful. The reader goroutine and ping goroutine are not stopped. I just fix the reader goroutine. If this PR is acceptable, we can fix the pinger goroutine in the same way.


alsm:
I think there is an issue with the change, I can't tell what you intended. If you can get it to work properly please resubmit a pr

PR #305, opened 2019-04-15
Use monotonic time for keep alive

### LINKED ISSUE #300: lastSent/keepAlive logic should not use wall clock
I happened to look through the source code and saw that ```lastSent```/```lastReceived``` uses epoch time. But the logic for keepAlive should really use time elapsed. Therefore I suggest that
```lastSent int64``` and ```lastReceived int64``` uses ```time.Time``` instead. Then when comparing times the monotonic clock will be used instead.

With the current logic the keep alive will break if the wall clock on the computer is changed.

Please advice if I have missunderstood something or if someone have a better solution.

johnhydroware:
resolves #300

### DIFF
--- a/client.go
+++ b/client.go
@@ -95,8 +95,8 @@ type Client interface {
 
 // client implements the Client interface
 type client struct {
-	lastSent        int64
-	lastReceived    int64
+	lastSent        atomic.Value
+	lastReceived    atomic.Value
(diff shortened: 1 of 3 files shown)


alsm:
Not sure why the check failed, I've validated there is an ECA

Issue #272, opened 2018-12-26
panic on reconnect


AhaEdgar:
WARN: 2018/12/26 11:53:00 Reconnection sucessful
2018-12-26T11:53:00.758819594Z panic: runtime error: invalid memory address or nil pointer dereference
2018-12-26T11:53:00.758841266Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x6e9e5e]
2018-12-26T11:53:00.758845334Z 
2018-12-26T11:53:00.758848408Z goroutine 8231 [running]:
2018-12-26T11:53:00.758851484Z mqtt-go/vendor/github.com/eclipse/paho%2emqtt%2egolang.(*client).internalConnLost(0xc000217880, 0xafd6e0, 0xc0304e4af0)
2018-12-26T11:53:00.758854779Z 	/go/src/mqtt-go/vendor/github.com/eclipse/paho.mqtt.golang/client.go:460 +0x5e
2018-12-26T11:53:00.758858029Z created by mqtt-go/vendor/github.com/eclipse/paho%2emqtt%2egolang.errorWatch
2018-12-26T11:53:00.758861003Z 	/go/src/mqtt-go/vendor/github.com/eclipse/paho.mqtt.golang/net.go:332 +0x1df

kbudde:
Hi,

I think I had the same issue. 
In my case it was caused because I triggered a (re)connect in the connectionLostHandler.
At the same time the Autoreconnect was trying to connect as well.

I think it is not a good idea to disturb the automatic reconnect.

zcxzcxczcx:
hi,i want to know you how to reconnect?

PR #310, opened 2019-04-23
when conn is nil,program will panic,so need to judge if the conn is nil.

gaohuag:
it can solve this issue:
https://github.com/eclipse/paho.mqtt.golang/issues/272


alsm:
Thanks for noticing this, I cannot merge PRs that don't have an ECA

PR #319, opened 2019-05-27
AddBroker: Escape % in Addresses

Mattemagikern:
The url.Parse(server) will fail if given an address containing a
non escaped % which is the case for an IPv6 address with a zone
indication.

Ex.
[fe80::a604:5838:8d8a:e532%wlp58s0]:1883 => [fe80::a604:5838:8d8a:e532%25wlp58s0]:1883

Regexp explanation:
`%(25)?`
% -- Match % literally (case sensitive)
(25)? --  Match 25 literally between zero and one time.

=> Will match any % followed by 25 or not.
The Regexp.ReplaceAll(...) will replace all matches of the regular
expression above and replace them with %25. Which is the URL escape code
for %.

Signed-off-by: Måns Ansgariusson <mansgariusson@gmail.com>

### DIFF
--- a/options.go
+++ b/options.go
@@ -10,6 +10,7 @@
  *    Seth Hoenig
  *    Allan Stockdill-Mander
  *    Mike Robertson
+ *    Måns Ansgariusson
  */
 
 // Portions copyright © 2018 TIBCO Software Inc.
@@ -20,6 +21,7 @@ import (
 	"crypto/tls"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


alsm:
Thanks

MattBrittan:
Thought I should add a note to this commit so relevant parties are notified. I am proposing to undo this change because it is causing issues elsewhere (when using web sockets the URL may contain % and escaping this breaks AWS auth). Given that RFC6874 states that % must be escaped I believe the best approach is to undo this change and leave it up to the caller to provide an address that complies with the standard. See issues #469 and #479 for further info.

MattBrittan:
This change has now been reversed in @master.

PR #322, opened 2019-06-20
Adding support for http/https proxy when using MQTT over websockets

reubenmiller:
## Summary

Switching websockets package to gorilla/websocket as it has support for http proxies and the package is maintained more frequently than golang.org/x/net/websocket.

When using either ws:// or wss:// connections, the http proxy settings can be set via the following standard environment variables.
* HTTP_PROXY
* HTTPS_PROXY
* NO_PROXY

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,11 @@ This client is designed to work with the standard Go tools, so installation is a
 go get github.com/eclipse/paho.mqtt.golang
 ```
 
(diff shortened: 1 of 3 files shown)


alsm:
Thanks for this, really useful.

iwittkau:
This is indeed really useful!
Are there any plans on releasing a new version including this PR @alsm?

PR #324, opened 2019-06-26
Add throttling option

johnhydroware:
Make it possible to throttle maximum number of messages sent per second.

### DIFF
--- a/net.go
+++ b/net.go
@@ -165,14 +165,33 @@ func incoming(c *client) {
 func outgoing(c *client) {
 	defer c.workers.Done()
 	DEBUG.Println(NET, "outgoing started")
+	var ticks <-chan time.Time
+	var throttleTicker *time.Ticker
+
+	obound, oboundP := c.obound, c.oboundP
+
+	if c.options.PublishThrottle > 0 {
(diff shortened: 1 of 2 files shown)


alsm:
I appreciate you trying to add this kind of function to the client, I know it would be useful for services like AWS IoT, but I'm not sure about this implementation. 1 message every 1/x seconds isn't quite the same as x messages per second.

johnhydroware:
> I appreciate you trying to add this kind of function to the client, I know it would be useful for services like AWS IoT, but I'm not sure about this implementation. 1 message every 1/x seconds isn't quite the same as x messages per second.

Would this functionality be accepted if I changed the implementation so that it was strictly x messages per second? If so, I will take a look at doing it that way.

alsm:
Yes, I was thinking how this might be done in the new v5 client, have you seen https://godoc.org/golang.org/x/time/rate this might help.

johnhydroware:
> Yes, I was thinking how this might be done in the new v5 client, have you seen https://godoc.org/golang.org/x/time/rate this might help.

That will limit the average rate, not the peak. The documentation for AWS doesn't really say what it means by "requests per second" and it it difficult to test since it will just silently discard messages. Unless I can't find any good documentation on this I will probably not do anything more on this issue.

Issue #334, opened 2019-07-11
"publish was broken by timeout" when WriteTimeout=0


MattBrittan:
I am getting "publish was broken by timeout" on approximately 10% of publish calls with the latest master. 

It appears that this is due to PR #326 which calls `time.After(c.options.WriteTimeout)` (client.go:604). If WriteTimeout is 0 (the default) then its possible that both options in the select statement will be ready so one will be chosen at random. Unfortunately when this happens the publish message is lost (I do not have code to deal with that because it should not be necessary with the filestore enabled).

Issue #325, opened 2019-07-01
Publishing freeze on connection lost


panter-dsd:
Version: master
If the option `options.AutoReconnect` is false then `options.MessageChannelDepth` is set to 0 [here](https://github.com/eclipse/paho.mqtt.golang/blob/0d81c295c555691f527a3e6f5b5c0a889f198bc3/client.go#L143). So, we have a channel with zero length. When we get into the line https://github.com/eclipse/paho.mqtt.golang/blob/0d81c295c555691f527a3e6f5b5c0a889f198bc3/client.go#L602 and connection is lost (channel reader has already exited), we're stuck at this line and never leave the `Publish` function.

ChIoT-Tech:
I believe that the fix introduces a new issue; if WriteTimeout is set to 0 (the default) then the "case <-time.After(c.options.WriteTimeout):" triggers immediately. This sometimes (approx 10% in my test code) ends up getting selected before the write to c.obound meaning that the publish message token is flagged as being in error state incorrectly (and the published message is lost despite AutoReconenct and persistence being active).

While setting WriteTimeout to a non-zero value would fix this the default value is currently is 0 (supposed to mean no timeout) and when used with AutoReconenct and Persistence these values make sense as the data will be written at some point. 
 
The simplest solution might be: 

```
if c.options.WriteTimeout == 0 {
	c.obound <- &PacketAndToken{p: pub, t: token}
} else {
	select {
	case c.obound <- &PacketAndToken{p: pub, t: token}:
	case <-time.After(c.options.WriteTimeout):
		token.setError(errors.New("publish was broken by timeout"))
	}
}
```

panter-dsd:
@ChIoT-Tech sorry, I didn't notice this. Fixed. Better would be to using contexts but this will change API.

MattBrittan:
Closing this because the change has been merged.

PR #335, opened 2019-07-13
Correcting behavior when WriteTimout is 0.

panter-dsd:
https://github.com/eclipse/paho.mqtt.golang/issues/325

Signed-off-by: Konstantin Konnov <konstantin.konnov@litmusautomation.com>


MattBrittan:
Hi @panter-dsd

Thanks for implementing a fix for this. Your approach is better than the one  I suggested (under the Chiot-Tech logon) because it means the Publish will not block for more than 30s (a good thing as the library is generally asynchronous).

closes #334 

Matt

alsm:
Thanks for this

PR #345, opened 2019-08-07
Allow using bytes.Buffer as payload

ctron:
It would be great if the Publish method could directly make use of a `bytes.Buffer`.


alsm:
Thanks for that, I made a slight update as your original didn't work.

PR #346, opened 2019-08-09
ADD SetForcePing

trevorseo:
It is to give an easy option for MQTT server to check online status of its clients. PINGREQ will be the only message the server to check every (e.g.) 30 seconds, out of (possibly) a lot of mqtt messages.


alsm:
Thanks for the idea but I'm not sure I see the value of sending extra pings.

PR #361, opened 2019-10-11
Persist outbound subscribe packets

### LINKED ISSUE #358: Outbound SubscribePackets are not persisted
I'm currently improving my application's resilience to network instability , and one of the edge cases I'm concerned with is sending subscription requests while the connection status is `connecting` or `reconnecting`.

I noticed that the `client.resume` method is designed to handle saved SubscribePackets if `options.ResumeSubs` is true. I've never seen this logic used, however, because it seems that outgoing SubscribePackets are not persisted. I would expect the `client.Subscribe` method would persist it's outgoing packets similar to how the `client.Publish` method persists outbound PublishPackets when the connection status is either `connecting` or `reconnecting`. the `persistOutbound` function also seems to contain the proper logic to save SubscribePackets.

I would love to add the logic to persist outbound subscribe packets in a new PR. Also, if my understanding of `ClientOptions.ResumeSubs` or the persistence of `SubscribePackets` is incorrect, I'm definitely interested in learning the correct behavior/usage so I can improve my application's design.

Thanks in advance!

robbawebba:
client: persist subscribe packets if client is connecting or reconnecting

The MQTT client has all of the logic in place to handle persisting
and resuming subscribe packets when the client is connecting
or reconnecting, but it's currently unused. This change persists
outbound SubscribePackets similar to Client.Publish() and adds an
fvt test case to validate the ResumeSubs behavior.

fixes #358

Signed-off-by: Robert Weber <robertweber95@gmail.com>


alsm:
Thanks for the work on this.

PR #363, opened 2019-10-16
Added support for MQTT protocols

spacanowski:
Some MQTT brokers use protocols `mqtt`, `mqtts` and `mqtt+ssl` (e.g. Amazon MQ).


alsm:
Thanks, I've added something similar, I couldn't accept this PR as it doesn't pass the ECA

PR #385, opened 2019-11-16
client: verify payload of SubscribeMultiple is not empty

### LINKED ISSUE #384: SubscribeMultiple accepts an empty map of topics
While I was debugging my application, I noticed that sending an empty map of topics/QoS to `client.SubscribeMultiple` does not result in an error, and the client will end up sending that SUBSCRIBE packet to the broker. I am currently working with AWS IoT's broker, and it considers this an error and closes the connection. The following gist shows an example that can trigger this behavior, along with the console output of running this example:
https://gist.github.com/robbawebba/7eebc964225fed7b10e61ac646b826bc

I'm not sure what the behavior is with other brokers. The MQTT 3.1.1 protocol section 3.8.3 specifies that the payload must contain at least 1 topic filter / QoS pair, so I'm guessing other brokers will not like this scenario either.

robbawebba:
MQTT protocol 3.1.1 specifies that the payload in a SUBSCRIBE packet
must not be empty and must contain at least one topic filter and QoS
pair. Validate the input to SubscribeMultiple meets this requirement to
avoid protocol errors when communicating with a broker.

Fixes #384

Signed-off-by: Robert Weber <robertweber95@gmail.com>

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -50,6 +50,10 @@ var ErrInvalidTopicMultilevel = errors.New("Invalid Topic; multi-level wildcard 
 //     Example:  a subscription to "foo/#" will match messages published to "foo".
 
 func validateSubscribeMap(subs map[string]byte) ([]string, []byte, error) {
+	if len(subs) == 0 {
+		return nil, nil, errors.New("Invalid subscription; subscribe map must not be empty")
+	}
+
 	var topics []string
 	var qoss []byte
 	for topic, qos := range subs {


client.go
```diff
@@ -754,6 +754,10 @@ func (c *client) SubscribeMultiple(filters map[string]byte, callback MessageHand
 			return token
 		}
 	}
+	if len(filters) == 0 {
+		token.setError(fmt.Errorf("topic filter/qos map must not be empty"))
+		return token
+	}
 	sub := packets.NewControlPacket(packets.Subscribe).(*packets.SubscribePacket)
 	if sub.Topics, sub.Qoss, err = validateSubscribeMap(filters); err != nil {
```
robbawebba: One idea I was unsure about was to add this error check to `validateSubscribeMap` because it might make more sense to group the topic/qos validation logic together. On the other hand, this is a simple check that could be performed as early as possible. Please let me know if you have a preference as to where this check should be performed and I'd be happy to make the adjustment.

alsm: If you could move this into the validateSubscribeMap option that would be great, subscribes aren't usually a frequent operation so moving this effectively a couple of lines further down should be fine and keeps the flow of the original

alsm:
Thanks

someburner:
@MattBrittan @alsm I also ran into this issue, but when I accidentally published to an empty topic `""`. Perhaps there should be a check in publish similar to this?

Issue #336, opened 2019-07-15
Unsubscribe surprisingly breaks previous subscribe


rollulus:
Subscribing to `/a/#` works correctly.
However, if one subscribes to `/a/b/#` and unsubscribes from it afterwards, the earlier message handler doesn't receive anything anymore, although the broker publishes according to `tcpdump`.

To reproduce:
```go
package main

import (
	"log"
	"time"

	mqtt "github.com/eclipse/paho.mqtt.golang"
)

func msgHandler(_ mqtt.Client, m mqtt.Message) {
	log.Printf("received message for topic %s: %s", m.Topic(), string(m.Payload()))
}

func main() {
	o := mqtt.NewClientOptions().AddBroker("tcp://" + "iot.eclipse.org:1883")
	c := mqtt.NewClient(o)
	if t := c.Connect(); t.Wait() && t.Error() != nil {
		panic(t.Error())
	}

	if t := c.Subscribe("/a/#", 1, msgHandler); t.Wait() && t.Error() != nil {
		panic(t.Error())
	}

	if t := c.Subscribe("/a/b/#", 1, msgHandler); t.Wait() && t.Error() != nil {
		panic(t.Error())
	}

	if t := c.Unsubscribe("/a/b/#"); t.Wait() && t.Error() != nil {
		panic(t.Error())
	}

	time.Sleep(time.Hour)
}
```

And publish messages from the console with:
```
mosquitto_pub -h iot.eclipse.org -t '/a/b/hello'   -d -m 'xxxx' -q 1
mosquitto_pub -h iot.eclipse.org -t '/a/hello'   -d -m 'xxxx' -q 1
```

My expectation is to receive those messages. In practice, nothing happens.

To verify that the example isn't broken itself, remove the last `Subscribe` and `Unsubscribe`.

boki:
I think this is the same issue I ran into: `router.addRoute` uses `route.match` to update existing routes, resulting in wildcards being applied. The fix seems to be to change [line 102](https://github.com/eclipse/paho.mqtt.golang/blob/master/router.go#L102) in route.go from

```diff
--- a/router.go
+++ b/router.go
@@ -105,7 +105,7 @@ func (r *router) addRoute(topic string, callback MessageHandler) {
        r.Lock()
        defer r.Unlock()
        for e := r.routes.Front(); e != nil; e = e.Next() {
-               if e.Value.(*route).match(topic) {
+               if e.Value.(*route).topic == topic {
[...]

PR #388, opened 2019-12-05
Fix callbacks being replaced on topics with wildcards

v-zhuravlev:
I had an issue with router.addRoute described here by fellow MQTT users: https://github.com/eclipse/paho.mqtt.golang/issues/336.
In my case, I have the client with subscriptions to `#`, `topic1`, `topic2` with different callbacks set for each. Due to `e.Value.(*route).match(topic)` (https://github.com/eclipse/paho.mqtt.golang/blob/master/router.go#L102) , new route/subscription overrides previous route instead of creating new one and only last callback is working for all routes.

And in my app I expect:
If I publish something to `topic1`, I expect both callbacks for `#` and `topic1` subscriptions to fire. Not just the last one.

### DIFF
--- a/router.go
+++ b/router.go
@@ -114,7 +114,7 @@ func (r *router) deleteRoute(topic string) {
 	r.Lock()
 	defer r.Unlock()
 	for e := r.routes.Front(); e != nil; e = e.Next() {
-		if e.Value.(*route).match(topic) {
+		if e.Value.(*route).topic == topic {
 			r.routes.Remove(e)
 			return
 		}
--- a/router.go
+++ b/router.go
@@ -99,7 +99,7 @@ func (r *router) addRoute(topic string, callback MessageHandler) {
 	r.Lock()
 	defer r.Unlock()
 	for e := r.routes.Front(); e != nil; e = e.Next() {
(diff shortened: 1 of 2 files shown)


MattBrittan:
Making this change and not also changing deleteRoute has the potential to be confusing (not that the existing code is not already confusing!). As the pull requestly currently stands calling:

```
Subscribe("#", xxx)
Subscribe("topic1", yyy)
Unsubscribe("topic1")
```

Would leave the route for "topic1" in place (Unsubscribe will call ```deleteRoute("topic1")``` which will match on "#", remove that route and return). The broker will continue to send all data through but only the callback for topic1 will remain active (definitly not what is intended).

The behaviour of the code in master as it currently stands is just as confusing; the above would result in no routes. However as this is a potentially breaking change to Subscribe then I believe that Unsubscribe/deleteRoute should also be updated to match at the same time. 

[Go Playground](https://play.golang.org/p/r20nvtq6fpd) demonstrating this.

v-zhuravlev:
Thanks @MattBrittan for such great comment, changed deleteRoute as well.

alsm:
Thanks for fixing this up.

PR #392, opened 2019-12-27
Add flag to mark if it is a dirty write 

dhananjaysathe:
Introduce DirtyWrite flag to indicate we did a no-op on write under certain conditions i.e c.connectionStatus() == reconnecting && qos == 0 so we can track this. 
This is useful in having more deterministic metrics and results. The alternate was erroring out with a new error like ErrDirtyWrite but that would break existing flows

### DIFF
--- a/client.go
+++ b/client.go
@@ -618,6 +618,7 @@ func (c *client) Publish(topic string, qos byte, retained bool, payload interfac
 		return token
 	case c.connectionStatus() == reconnecting && qos == 0:
 		token.flowComplete()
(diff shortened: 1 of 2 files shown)


MattBrittan:
Sorry - its not possible to commit this without the necessary legal agreements (see the eclipsefdn/eca check); signing the agreement is fairly simple and protects users of the package.

MattBrittan:
Closing this due to lack of ECN and inactivity.

Issue #356, opened 2019-09-29
Broken TLS 1.3 handshake error handling


akokhanovskyi:
When a broker rejects the client cert, the expected token error from `Client.Connect()` is `Network Error : remote error: tls: bad certificate`. The actual error with TLS 1.3 is `Network Error : %!s(<nil>)`. TLS 1.3 is the [default with go 1.13; opt-out will be removed in go 1.14](https://golang.org/doc/go1.13#tls_1_3).

Why this happens:

> In TLS 1.3 the client is the last one to speak in the handshake, so if it causes an error to occur on the server, it will be returned on the client by the first Read, not by Handshake. For example, that will be the case if the server rejects the client certificate.

(see [go 1.12 release notes](https://golang.org/doc/go1.12#tls_1_3))

According to the above:

- `tls.DialWithDialer()` yields no error (see [`net.go:76`](https://github.com/eclipse/paho.mqtt.golang/blob/master/net.go#L76))
- `openConnection()` also returns a `nil` error ([`client.go:255`](https://github.com/eclipse/paho.mqtt.golang/blob/master/client.go#L255))
- `client.connect()` fails with return code `packets.ErrNetworkError` ([`client.go:281`](https://github.com/eclipse/paho.mqtt.golang/blob/master/client.go#L281))
- `token.setError()` sets inaccurate error ([`client.go:324`](https://github.com/eclipse/paho.mqtt.golang/blob/master/client.go#L324)): `err` is still `nil`, which leads to an `fmt` error (`%!s(<nil>)`).

I think, this can be fixed by properly inspecting the error returned by `packets.ReadPacket()` ([`client.go:478`](https://github.com/eclipse/paho.mqtt.golang/blob/master/client.go#L478)). In case of a rejected client cert, this is a `net.OpError` remote error with `alertBadCertificate`. I'm not well-versed in the overall library design, though, so maybe there are better ideas.

Dunken:
>In case of a rejected client cert, this is a `net.OpError` remote error with `alertBadCertificate`.

Did you verify this? I always get a `alertDecryptError`.

akokhanovskyi:
I tested with an intentionally wrong client-presented x.509 cert that was not trusted by the server CA. Your error seems to be different: `alertDecryptError` normally means that the server is somehow unable to decrypt the first encrypted message from the client ("Finished").

akokhanovskyi:
@alsm Could you please review #452? Looking forward to your feedback!

PR #404, opened 2020-03-06
fix(356): return string if read after TLS handshake fails

Dunken:
In TLS 1.3 the client is the last one to speak in the handshake, so if it causes an error to occur on the server, it will be returned on the client by the first Read, not by Handshake. For example, that will be the case if the server rejects the client certificate.

Bug: https://github.com/eclipse/paho.mqtt.golang/issues/356
Signed-off-by: Armin Galliker <mc_ghc@yahoo.de>

### DIFF
--- a/client.go
+++ b/client.go
@@ -282,7 +282,7 @@ func (c *client) Connect() Token {
 				}
 				cm.Write(c.conn)
 
-				rc, t.sessionPresent = c.connect()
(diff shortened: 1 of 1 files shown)


MattBrittan:
Closing because I believe that this is addressed in PR #452 (which updated your PR to reflect changes in @master). Thanks for pointing this issue out and raising a PR.

PR #405, opened 2020-03-09
Update options.go

ackratos:


alsm:
Thanks for spotting this, unfortunately I cannot merge as the ECA does not pass, I will update the doc lines.

PR #408, opened 2020-03-12
no more uncatched errors

thomas-tacquet:
I think these errors should be catched or at least logged


alsm:
Thanks for this, I've made similar changes, I could not accept the PR as it does not pass the ECA

PR #414, opened 2020-04-30
Add websocket options

smitt04:
I added WebSocket options as I needed to change the default WriteBufferSize. I could not send payloads larger than 4096 bytes because the server did not support continuation frames. By setting the `WriteBufferSize` to 128k, the largest supported payload size for my server I was able to send large requests again.


smitt04:
I am not sure how to get the eca check fixed. I have signed the agreement, originally with a different email address, and now it fails.

alsm:
The ECA needs to be signed by the same email as is on the commit, I think it's a simple email lookup

smitt04:
Yes, and it is now. But when I click details it says the validation request is not available

alsm:
I'm sorry but it doesn't give me anything else or any options either. You could try closing this PR and opening a new one, see if that kicks it into working.

smitt04:
Yea I tried that once the first day. I can try closing it again and submit a new one on Monday.

PR #417, opened 2020-05-04
Add websocket options

smitt04:
I added WebSocket options as I needed to change the default WriteBufferSize. I could not send payloads larger than 4096 bytes because the server did not support continuation frames. By setting the `WriteBufferSize` to 128k, the largest supported payload size for my server I was able to send large requests again.


smitt04:
Will resubmit with commits reorganized

PR #420, opened 2020-05-07
corrects panic on 'CONNACK was not CONN_ACCEPTED'

MattBrittan:
Reverted back to code previously here, was changed in error (in my big net refactor I believe).
closes 419


d21d3q:
Works fine after this change.

PR #431, opened 2020-06-18
Add Proxy to WebsocketOptions

fulder:
Add support for custom proxy function for websocket Dialer instead of hardcoding it to `http.ProxyFromEnvironment`. This hardcode can cause unexpected behavior during use cases where the proxy is changed during runtime as `http.ProxyFromEnvironment` will only look it up once during the first call.

Also added a warning log if websocket handshake fails due to e.g. invalid status code (other than 101). This response slurped and returned by the `Dial` function was just ignored and made debugging harder.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -47,10 +47,10 @@ func NewWebsocket(host string, tlsc *tls.Config, timeout time.Duration, requestH
 
 	ws, resp, err := dialer.Dial(host, requestHeader)
 
-	if resp != nil {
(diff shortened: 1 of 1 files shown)


jeolax:
No activity in nine months. This PR would be helpful when proxy isn't set with HTTP_PROXY environment variables. How ca we bring this PR forward?

MattBrittan:
Apologies - this came through before I became a committer and I did not notice it when reviewing outstanding commits. Looks good to me.

fulder:
@MattBrittan no worries, thank you for the review and merge :+1:

jeolax:
Great! Thanks for the quick response @MattBrittan

PR #447, opened 2020-08-07
Add Done() method to Token for use with select

### LINKED ISSUE #445: Add method to Token interface for getting pull-only channel
The existing `Wait` and `WaitTimeout` methods on the `Token` interface are fairly inflexible. Since these are blocking calls, there is no way to stop waiting other than for it to succeed, fail, or timeout. For example, if I want to stop waiting upon context cancellation, there is no way to do it without leaking a goroutine.

```go
result := make(chan bool, 1)
go func() {
    result <- token.WaitTimeout(timeout)
}()

select {
case ok := <-result:
    ...
case <-ctx.Done():
    // There is no way to "abort" the goroutine above.
   return ctx.Err()
}
```

Instead, the `Token` interface should have a method for fetching the token's completion channel, which is closed when the operation in question succeeds or fails.

```go
type Token interface {
    ...
    Done() <-chan struct{}
}

ctx, cancelFunc := context.WithTimeout(ctx, timeout)
defer cancelFunc()

select {
case <-token.Done():
    return token.Error()
case <-ctx.Done():
    return ctx.Err()
}
```

rittneje:
Add a `Done` method to the `Token` interface so we can abort waiting upon context cancellation, for example. Fixes #445.

### DIFF
--- a/messageids.go
+++ b/messageids.go
@@ -98,12 +98,21 @@ type DummyToken struct {
 	id uint16
 }
 
+// Wait implements the Token Wait method.
 func (d *DummyToken) Wait() bool {
 	return true
 }
 
+// WaitTimeout implements the Token WaitTimeout method.
 func (d *DummyToken) WaitTimeout(t time.Duration) bool {
 	return true
+}
+
+// Done implements the Token Done method.
+func (d *DummyToken) Done() <-chan struct{} {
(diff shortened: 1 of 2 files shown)


rittneje:
@alsm Please take a look.

alsm:
Thanks for the contribution

PR #448, opened 2020-08-13
Add cookie jar option to WebsocketOptions

nejtr0n:
Adding http.CookieJar option to pass websocket client


MattBrittan:
Sorry for not responding to this PPR sooner; unfortunately because the PR includes commits from two users (well email addresses) and only one of these is covered by the Eclipse Foundation Contributor Agreement I'm unable to accept it. As its a very simple change I think the simplest approach is for me to close this PR and, if you wish, you can resubmit as one commit using an address covered by the ECA.

PR #452, opened 2020-09-17
Propagate MQTT connect error (fixes #356)

### LINKED ISSUE #356: Broken TLS 1.3 handshake error handling

akokhanovskyi:
PR #404 is in conflict with the master, so this is a new fix for #356. Along with accurately propagating an MQTT connect error, multiple [golangci-lint](https://github.com/golangci/golangci-lint) errors are fixed (only in the `net.go` to reduce the PR size).

Signed-off-by: Andrew Kokhanovskyi a.kokhanovskyi@gmail.com

### DIFF
--- a/client.go
+++ b/client.go
@@ -365,7 +365,7 @@ func (c *client) attemptConnection() (net.Conn, byte, bool, error) {
 		DEBUG.Println(CLI, "socket connected to broker")
 
 		// Now we send the perform the MQTT connection handshake
-		rc, sessionPresent = ConnectMQTT(conn, cm, protocolVersion)
+		rc, sessionPresent, err = connectMQTT(conn, cm, protocolVersion)
 		if rc == packets.Accepted {
 			break // successfully connected
 		}
--- a/net.go
+++ b/net.go
@@ -15,6 +15,8 @@
 package mqtt
 
 import (
+	"errors"
+	"io"
 	"net"
 	"reflect"
 	"strings"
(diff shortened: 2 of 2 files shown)


akokhanovskyi:
Updated to keep up to speed with the current master

alsm:
Thanks for the work on this.

PR #455, opened 2020-10-02
Point documentation to pkg.go.dev

damienstanton:
Most projects have moved away from godoc.org as the standard hosting site for generated documentation. This PR updates the reference link and badge, accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-
-[![GoDoc](https://godoc.org/github.com/eclipse/paho.mqtt.golang?status.svg)](https://godoc.org/github.com/eclipse/paho.mqtt.golang)
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/eclipse/paho.mqtt.golang)](https://pkg.go.dev/github.com/eclipse/paho.mqtt.golang)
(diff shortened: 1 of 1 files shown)


damienstanton:
Note that I just signed an ECA, so this might require a manual trigger to get that check to pass.

MattBrittan:
Note: Before doing this we need to look at how the license info is stored. Currently it appears that google does not detect the licenses ([as seen here](https://pkg.go.dev/github.com/eclipse/paho.mqtt.golang@master)) meaning that the documentation will not be visible; this is likely to become an issue when we make the next release (if this commit is merged).

@damienstanton - unfortunately to pass the automated test the "Signed-off-by" entry ([more info here](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git)) needs to be on all commits. Signing out the offending commit and using `git commit --amend --signoff` seemed to work when I had this issue.

MattBrittan:
I have significantly updated the README (#461) and to complete this have just raised #462 which makes the change you raised as well as a few others:
 * Changes the link further down the document
 * Updates the LICENSE file so that pkg.go.dev will display the documentation when the next version is released.

Thanks very much for your contribution and apologies for duplicating it in another request (hopefully we can release v1.3.0 shortly so I wanted to test that the documentation links will continue to function).

MattBrittan:
Closing this PR because the changes were implemented in PR #462 (which made further changes to avoid issues with the license file). Thanks for pointing out the issue and raising a PR.

PR #461, opened 2020-10-31
Update README and add docker example (mainly as a base for reproducing issues)

MattBrittan:
Having reviewed current issues, and considered questions raised elsewhere, I feel that we need to provide more information to assist those starting out with the library and those experiencing issues (particularly in regards to the level of information needed when raising an issue).

This pull request significantly updates/enhances the `readme.md` and adds `cmd/docker` (which should be a useful example as well as providing a good base for those wanting to produce a reproducible example of an issue). Note: I considered renaming `cmd` to `examples` to make it more discoverable but have left it as-is for now.

A key aim of this change is to provide something that can be referred to when replying to issues that provide insufficient information (removing the need to copy/paste from, or provide links to, other issues).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,24 +6,39 @@ Eclipse Paho MQTT Go client
 ===========================
 
 
-This repository contains the source code for the [Eclipse Paho](http://eclipse.org/paho) MQTT Go client library. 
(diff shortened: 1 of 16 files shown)


alsm:
Thanks for adding all this

PR #475, opened 2021-01-03
Export connection refused errors outside of the map

umeat:


umeat:
It's really annoying to have to access these errors inside the map. The code is functionally exactly the same after this change, but users can directly references these errors.

Happy to change the names if anyone thinks that needs to be done.

MattBrittan:
Thanks for the pull request; makes sense to me.

PR #492, opened 2021-03-22
add delay in reconnect to prevent 100% cpu on re-connect issues

rdoorn:


MattBrittan (changes requested):
This would prevent the reconnection from ever succeeding (the `if err == nil {` is how the loop is exited on success; after your modification it cannot succeed). Please raise (or reopen if one has been inappropriately closed) an issue detailing the problem you are trying to solve with this.

PR #494, opened 2021-03-31
Log response code if websocket dial fails.

### LINKED ISSUE #493: Improve error reporiting in NewWebsocket
NewWebsocket return error "websocket: bad handshake" without any details.
Gorila websocket return error "websocket: bad handshake" and  details are in response object.
NewWebsocket is omitting response object and thus hidding cause of error.
https://github.com/eclipse/paho.mqtt.golang/blob/18bfbdece2e98c020293755a3468bf510a7b2497/websocket.go#L41
Would it be possible to get more details? 

	ws, res, err := dialer.Dial(host, requestHeader)

	if err != nil {
		if res!=nil{
			DEBUG.Println(CLI,res.Status)
		}
		return nil, err
	}

MattBrittan:
closes #493

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -38,16 +38,18 @@ func NewWebsocket(host string, tlsc *tls.Config, timeout time.Duration, requestH
 		WriteBufferSize:   options.WriteBufferSize,
 	}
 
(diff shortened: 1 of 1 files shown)


MattBrittan:
This has already been done!

Issue #496, opened 2021-04-22
Add HTTP Connect proxy support


amir-khassaia:
Raising this as a request to support connections via HTTP forwarding/tunneling proxies in addition to SOCKS proxies that are already natively supported. It seems that other paho flavours already have support for this too.

It's possible to work with proxies better by supporting connection tunneling offered by HTTP/CONNECT for compatible proxies as defined in https://tools.ietf.org/html/rfc7231#section-4.3.6.

This works by having the compliant proxy transparently relaying the data and the client issues the connect just writes the data to the proxy without needing to know much more.

Can submit a PR, I've not yet signed an ECA as a contributor though.

Related to #394

MattBrittan:
Please feel free to submit a PR, however it will not be possible to accept it until you have signed the ECA (it's a fairly simple process and is important in order to ensure the code is appropriately licensed).

amir-khassaia:
@MattBrittan: I've created an account and signed an ECA against the same github account as I'm pushing from. 
However I'm not able to push a new branch to the remote with changes for PR

> remote: Permission to eclipse/paho.mqtt.golang.git denied to amir-khassaia.

MattBrittan:
@amir-khassaia you will not be able to push a branch to this repo; the standard process is to create a fork and [generate the pull request from that](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork)

amir-khassaia:
PR done but ECA doesn't appear to be recognized - I've signed and resigned the ECA and double checked the emails match. Will revisit again later.

MattBrittan:
I can see that you have signed the ECA - there have been some changes made to the ECA verification recently (sign-off within comment no longer needed) so it's possible that this is broken (I cannot see anything obviously wrong). Given that I have no issues with the ECA but have made a few comments (am concerned about adding more complexity into this package, especially when there are side effects like in `http_proxy.go`. As I see it the user could do this in their own code (we could add a demo that includes this) and that avoids adding it here.

amir-khassaia:
Agree, I've left comments along the similar thinking in the PR too: https://github.com/eclipse/paho.mqtt.golang/pull/497#discussion_r618059448
All this should become is:
-  having the proxy example included 
-  A new callback that will be executed prior to making the TLS handshake to allow SNI injection
The above will satisfy all the use cases whilst keeping the library clean.

amir-khassaia:
[...]

PR #497, opened 2021-04-22
Feat/http connect proxy support

amir-khassaia:
See #496 
- Support for HTTP CONNECT proxy via environment variable
- Improved TLS support for SNI extension to be emitted

### DIFF
--- a/http_proxy.go
+++ b/http_proxy.go
@@ -1,0 +1,85 @@
+package mqtt
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/proxy"
+)
+
+// httpProxy is a HTTP/HTTPS connect capable proxy.
+type httpProxy struct {
+	host     string
+	haveAuth bool
+	username string
+	password string
+	forward  proxy.Dialer
+}
+
+func (s httpProxy) String() string {
+	return fmt.Sprintf("HTTP proxy dialer for %s", s.host)
+}
+
+func newHTTPProxy(uri *url.URL, forward proxy.Dialer) (proxy.Dialer, error) {
+	s := new(httpProxy)
+	s.host = uri.Host
+	s.forward = forward
+	if uri.User != nil {
+		s.haveAuth = true
+		s.username = uri.User.Username()
+		s.password, _ = uri.User.Password()
+	}
+
+	return s, nil
+}
+
+func (s *httpProxy) Dial(_, addr string) (net.Conn, error) {
+	reqURL := url.URL{
+		Scheme: "https",
+		Host:   addr,
+	}
+
+	req, err := http.NewRequest("CONNECT", reqURL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Close = false
+	if s.haveAuth {
+		req.SetBasicAuth(s.username, s.password)
+	}
+	req.Header.Set("User-Agent", "paho.mqtt")
+
+	// Dial and create the client connection.
+	c, err := s.forward.Dial("tcp", s.host)
+	if err != nil {
+		return nil, err
+	}
+
+	err = req.Write(c)
+	if err != nil {
+		_ = c.Close()
+		return nil, err
+	}
+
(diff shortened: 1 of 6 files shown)


http_proxy.go
```diff
@@ -0,0 +1,84 @@
+package mqtt
```
MattBrittan, line 1: Not sure if it's best to include this (httpProxy) in the library or get the user to do it in their code? (it does not look there is anything stopping them). Doing it here complicates changes if the user has specific requirements not met by this code; perhaps it would be better to create a proxy example that includes this?

netconn.go
```diff
@@ -77,7 +77,7 @@ func openConnection(uri *url.URL, tlsc *tls.Config, timeout time.Duration, heade
 			return nil, err
 		}
 
-		tlsConn := tls.Client(conn, tlsc)
+		tlsConn := tls.Client(conn, tlsConfigWithSni(uri, tlsc))
```
MattBrittan, line 80: Fine with this - I don't think it will break existing code. However I am going to have a think about the approach because I wonder if using a callback (say `func OnConnectionAttempt(broker *url.URL, tlsCfg *tls.Config) *tls.Config`) might provide more flexibility?

http_proxy.go
amir-khassaia: No, nothing stopping them its just going to be more code to maintain by each consumer - however I can see the argument for this being not the responsibility of the library, in which case an example might suffice too.

netconn.go
amir-khassaia: Didn't see a reference to this callback on clone of master, if this is more aligned with idiomatic use of the library then it fits better (this can then be an SNI aware connection callback impl) - as long as it's done prior to the handshake should be fine - which looking at the name of it should be the case.

MattBrittan: Its not something that currently exists (would be a new callback). However it would solve solve some other issues too (currently there is only one *tls.Config but you can specify multiple brokers to attempt connections to).

amir-khassaia: I think it might be a good idea to have such a call back exposed (`func OnConnectionAttempt`) and then we can decline the PR and not need this explicit change, allowing the consumer again to set the SNI in the callback as needed, keeping the library clean.
So I think all thats needed is an example of the proxy and a callback and the PR can then be closed allowing other consumers to implement this functionality using those tools.

MattBrittan: I'm happy to add the callback (might be faster as this will touch a number of files). Are you happy to do a demo? (I don't use proxies so it's not something I can really add!).

amir-khassaia: Oh I think we got our wires crossed, I've already done the callback unless you think I missed something.
I've used the example app I've included to test against mqtt.googleapis.com with a custom proxy on localhost

options.go
```diff
@@ -321,6 +326,13 @@ func (o *ClientOptions) SetReconnectingHandler(cb ReconnectHandler) *ClientOptio
 	return o
 }
 
+// SetConnectionAttemptHandler sets the ConnectionAttemptHandler callback to be executed prior
```
MattBrittan, line 329: Think this comment should be something along the lines of:
// SetConnectionAttemptHandler sets the ConnectionAttemptHandler callback to be executed prior
// to each attempt to connect to an MQTT broker. Returns the *tls.Config that will be used when establishing
// the connection (a copy of the tls.Config from ClientOptions will be passed in along with the broker URL). This
// allows connection specific changes to be made to the *tls.Config.

MattBrittan (commented):
Looks good - would just like a little more info in the comment (as that forms the documentation)

MattBrittan:
For future reverence: Verified that @amir-khassaia has signed the ECA (https://accounts.eclipse.org/legal/eca along with the @gmail.com email address in the signed-off footer). I am satisfied that the author of the commit has signed the ECA.

options.go
amir-khassaia: Thanks, pushed updated comment

MattBrittan:
Thanks very much for your contribution.

amir-khassaia:
> Thanks very much for your contribution.

Happy to contribute, I will probably end up consuming once this is released in the future 😉

PR #540, opened 2021-09-20
Use uri.Path instead of uri.Host for UNIX domain socket URLs

webconn:
This patch allows to use proper UNIX socket URIs such as `unix:///var/run/socket.sock`. This format is OK with RFC3986 and used in [Docker](https://github.com/docker/cli/blob/8e08b72450719baed03eed0e0713aae885315bac/man/dockerd.8.md), for example.

Current implementation makes possible only relative paths (like `unix://socket.sock`) and it looks like it does so by accident.

Example of how it is parsed by `url.Parse()`: https://play.golang.org/p/OajK5IJX2F9

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	val, err := url.Parse("unix:///var/run/socket.sock")
	fmt.Printf("%#v %v", val, err)
}
```

```
&url.URL{Scheme:"unix", Opaque:"", User:(*url.Userinfo)(nil), Host:"", Path:"/var/run/socket.sock", RawPath:"", ForceQuery:false, RawQuery:"", Fragment:"", RawFragment:""} <nil>
```
```

### DIFF
--- a/netconn.go
+++ b/netconn.go
@@ -62,7 +62,7 @@ func openConnection(uri *url.URL, tlsc *tls.Config, timeout time.Duration, heade
 		}
 		return conn, nil
 	case "unix":
-		conn, err := net.DialTimeout("unix", uri.Host, timeout)
+		conn, err := net.DialTimeout("unix", uri.Path, timeout)
 		if err != nil {
 			return nil, err
 		}
--- a/netconn.go
+++ b/netconn.go
@@ -62,7 +62,17 @@ func openConnection(uri *url.URL, tlsc *tls.Config, timeout time.Duration, heade
 		}
 		return conn, nil
 	case "unix":
-		conn, err := net.DialTimeout("unix", uri.Path, timeout)
+		var conn net.Conn
+		var err error
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by MattBrittan** [comment]
Unfortunately it will not be possible to merge this until you sign the ECA (see the [contributing](https://github.com/eclipse/paho.mqtt.golang#contributing) section of the readme).

This is not a feature I have used (and I don't really know all that much about the UNIX socket interface) but I am concerned that this may break existing uses (e.g. the example you gave ```unix://socket.sock```). I may be wrong but assume whoever added this functionality had a use for it the way its written; would something like the following work (without breaking any existing usage):

```
var conn net.Conn
var err error
if len(uri.Host) > 0 {
    conn, err = net.DialTimeout("unix", uri.Host, timeout)
else {
    conn, err = net.DialTimeout("unix", uri.Path, timeout)
}
```

webconn:
Good point, added `uri.Host` check for compatibility. Also I signed the ECA just now, may require triggering checks.

MattBrittan:
Looks good to me; thanks for your contribution.

PR #541, opened 2021-09-22
CodeQL & integer conversion fix

jfcg:


MattBrittan (approved):
I guess this change is technically correct (the chance of the existing conversion causing an issue is very slight - a file would need to be manually added to the store folder).

PR #558, opened 2021-11-10
log err.error() if ConnectRetry is enabled

TomasVojacek:
If ConnectRetry is enabled and connection fail, it is hard to determine root cause of failure.
This is change will print error to log.

### DIFF
--- a/client.go
+++ b/client.go
@@ -282,7 +282,7 @@ func (c *client) Connect() Token {
 		conn, rc, t.sessionPresent, err = c.attemptConnection()
 		if err != nil {
 			if c.options.ConnectRetry {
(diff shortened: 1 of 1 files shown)


MattBrittan:
Thanks for the contribution; unfortunately I cannot accept it without a [signed ECA](https://github.com/eclipse/paho.mqtt.golang#contributing) (this ensures that contributions are appropriately licenced; I realise this seems like overkill for such a small change but unfortunately its a requirement for any change).

TomasVojacek:
ECA signed

TomasVojacek:
> Thanks for the contribution; unfortunately I cannot accept it without a [signed ECA](https://github.com/eclipse/paho.mqtt.golang#contributing) (this ensures that contributions are appropriately licenced; I realise this seems like overkill for such a small change but unfortunately its a requirement for any change).

ECA signed

PR #561, opened 2021-11-12
Add tcp option to set keep alive value on the tcp layer

mogaleaf:
Hi,

Our devices have bandwidth constraint and we can't use the default 15s tcp keep-alive from the go sdk library.
I have set up an option to be able to change it as it is impossible to do so with the library.

Let me know if it could be integrated to the library.

Signed-off-by: cecile <cecile.thiebaut@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -393,7 +393,7 @@ func (c *client) attemptConnection() (net.Conn, byte, bool, error) {
 			tlsCfg = c.options.OnConnectAttempt(broker, c.options.TLSConfig)
 		}
 		// Start by opening the network connection (tcp, tls, ws) etc
-		conn, err = openConnection(broker, tlsCfg, c.options.ConnectTimeout, c.options.HTTPHeaders, c.options.WebsocketOptions)
+		conn, err = openConnection(broker, tlsCfg, c.options.ConnectTimeout, c.options.HTTPHeaders, c.options.WebsocketOptions, c.options.TcpOptions)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


MattBrittan:
Thanks for this; I can see why it might be useful. However I'm not completely comfortable with the implementation due to the added complexity (extra config structure `TcpOptions` etc) and if we are allowing the user to configure this then why not other `Dialer` options (`LocalAddr` and `FallbackDelay` could be useful).

I wonder if it would be simpler to add a `Dialer net.Dialer` in `ClientOptions` (default to `net.Dialer{}` and have `SetConnectTimeout` manipulate that). You could then add a `SetDialer` that provides a fully custom `Dialer`.  The `Dialer` could then be passed into `openConnection` (simplifying that function). This would mean importing `net` into `options.go` but as this already imports a couple of net packages I don't see that being an issue.

mogaleaf:
I have implemented the changes using net.Dialer directly into options.
Set up default timeout to 30 * time.Second for backward compatibility.
ConnectTimeout is used for WebSocket too so I need to keep it at options level and dialer level.
Let me know what you think about this option.

Thanks

MattBrittan:
Thanks very much,
As you say having the options `ConnectTimeout` and `Dialer` is not ideal but I think working around that is likely to introduce further confusion (the alternative being to pass `dialer.Timeout` to `openConnection`). The way this is implemented now should provide maximum flexibility with minimal code changes. This passes the tests so I'll accept it.

Matt

Issue #563, opened 2021-12-03
token.Wait() does not wait for completion 


ecksun:
I'm trying to use this library in a test. What I want to do is send some messages and then exit. My broker will disconnect the mqtt connection if I try to send messages too early so I want to retry until I know the message is delivered so I can exit when that is done.

The issue is that `token.Wait()` doesn't always wait until the message is actually delivered. Reading through issues, especially [this](https://github.com/eclipse/paho.mqtt.golang/pull/450#issuecomment-684042034) comment from @MattBrittan it seems like this is "intended" behavior.

Here is some code to reproduce the issue:

```
package main

import (
	"fmt"
	"log"
	"os"

	mqtt "github.com/eclipse/paho.mqtt.golang"
)

func main() {
	mqtt.ERROR = log.New(os.Stdout, "[ERROR] ", 0)
	mqtt.CRITICAL = log.New(os.Stdout, "[CRIT] ", 0)
	mqtt.WARN = log.New(os.Stdout, "[WARN]  ", 0)
	mqtt.DEBUG = log.New(os.Stdout, "[DEBUG] ", 0)
	opts := mqtt.NewClientOptions().
		AddBroker("tcp://localhost:1883").
		SetClientID("tester").
		SetUsername("hunter")

	c := mqtt.NewClient(opts)
	if token := c.Connect(); token.Wait() && token.Error() != nil {
		panic(token.Error())
	}

	token := c.Publish("test", 1, false, []byte{})
	token.Wait()
	if token.Error() != nil {
		fmt.Printf("token.Error() = %+v\n", token.Error())
	}

	c.Disconnect(250)
}
```

and the logs:

```
[DEBUG] [client]   Connect()
[DEBUG] [store]    memorystore initialized
[DEBUG] [client]   about to write new connect msg
[DEBUG] [client]   socket connected to broker
[DEBUG] [client]   Using MQTT 3.1.1 protocol
[DEBUG] [net]      connect started
[DEBUG] [net]      received connack
[DEBUG] [client]   startCommsWorkers called
[DEBUG] [client]   client is connected/reconnected
[DEBUG] [net]      incoming started
[DEBUG] [net]      startIncomingComms started
[DEBUG] [net]      outgoing started
[DEBUG] [net]      startComms started
[DEBUG] [pinger]   keepalive starting
[DEBUG] [client]   startCommsWorkers done
[DEBUG] [net]      outgoing waiting for an outbound message
[...]

PR #565, opened 2021-12-06
Document token.Wait behaviour during reconnect

ecksun:
When a connection is lost to the broker and later resumed the token
returned to the user will be considered completed and a new token will
be created instead. The result of this is that there is no way for a
user of this library to know whether a message has been delivered in the
presence of connection issues.

This PR documents the behavior found in https://github.com/eclipse/paho.mqtt.golang/issues/563

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,9 @@ configure a handler with `AddRoute` or set a `DefaultPublishHandler`.
 * Reusing a `Client` is not completely safe. After calling `Disconnect` please create a new Client (`NewClient()`) rather 
 than attempting to reuse the existing one (note that features such as `SetAutoReconnect` mean this is rarely necessary).
 * Brokers offer many configuration options; some settings may lead to unexpected results.
+* Publish tokens will complete if the connection is lost and re-established using the default
+options.SetAutoReconnect(true) functionality (token.Error() will return nil). Attempts will be made to re-deliver the
+message but there is currently no easy way know when such messages are delivered.
 
 If using Mosquitto then there are a range of fairly common issues:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -129,6 +129,7 @@ may lead to messages being silently dropped.
 * `persistence` - Defaults to false (messages will not survive a broker restart)
 * `max_keepalive` - defaults to 65535 and, from version 2.0.12, `SetKeepAlive(0)` will result in a rejected connection 
 by default.
+* `token.Wait()` - Will return successfully before a message has been delivered if the client needs to reconnect to the broker before the message has been delivered.
```
MattBrittan, line 132: Move up above the "If using Mosquitto then there are a range of fairly common issues:" line (as this is not specific to Mosquitto). I would suggest something like the following (to mirror the format of previous comments and reflect the impact of options):

Publish tokens will complete when the connection is lost (in most cases `token.Error()` will return `nil` - see the `internalConnLost` function in `client.go` for full details). The message may be delivered to the broker when the connection is re-established (either via the `AutoReconnect` functionality or manually depending upon the `CleanSession`/`Store` `ClientOptions`)

token.go
```diff
@@ -37,12 +37,18 @@ type PacketAndToken struct {
 type Token interface {
 	// Wait will wait indefinitely for the Token to complete, ie the Publish
 	// to be sent and confirmed receipt from the broker.
+	//
```
MattBrittan, line 40: I'm not sure if we want to add this to the interface definition because this formalises the behaviour. Currently it's undocumented which leaves some room to change it (which probably should probably happen - for the reconnect anyway - its just a matter of someone putting the time in to make the change).

PR #568, opened 2021-12-15
Add SimulateConnectionLoss interface to help with testing registered callbacks

gangabhavanik:
This interface simply calls existing internal interface internalConnLoss.

Testing: I have tested it by running mosquitto mqtt broker and running 'go test'

```
$ go test 
...
connection lost before Publish completed
choke is waiting
choke is waiting
i2: 2
PASS
ok      github.com/eclipse/paho.mqtt.golang     25.786s
```

I am adding this interface to give me the ability to unit test OnConnectHandler callback code in the project that I am currently working on.

### DIFF
--- a/client.go
+++ b/client.go
@@ -118,6 +118,9 @@ type Client interface {
 	// OptionsReader returns a ClientOptionsReader which is a copy of the clientoptions
 	// in use by the client.
 	OptionsReader() ClientOptionsReader
+	// SimulateConnectionLoss helps in tesing callbacks that are registered when clients are
+	// created with AutoReconnect enabled.
+	SimulateConnectionLoss(err error)
 }
 
 // client implements the Client interface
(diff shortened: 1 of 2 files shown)


gangabhavanik:
Let me close this and submit with id that I have registered on eclipse.

MattBrittan:
Before you resubmit this please consider what benefits this delivers and state them in the pull description (or, ideally, open an issue first stating the issue you are trying to address so we can discuss). Changing this to `SimulateConnectionLoss` adds a new externally accessible function (`internalConnLost` is private). 

`internalConnLost` was implemented for testing only (hence the internal in the name). As a result the function itself has not been thoroughly tested and using it may well lead to deadlocks etc. Due to the way this library is written (heaps of options) it is difficult to predict how its used and very easy to introduce additional deadlocks when exposing new functionality. As such it's important that there is a clear rationale for adding functionality and it needs to be well tested.

gangabhavanik:
Thank you. I'll open an issue stating what I am trying to solve.

PR #569, opened 2021-12-16
add cuscom dialer

carr123:
hi, i add cuscom dialer to support more underlying transport types.

it's easy to use. 

demo :

import (
     mqtt "github.com/eclipse/paho.mqtt.golang"
)

func main() {
	mqtt.AddCustomDialer(
		"haha",
		func(uri *url.URL, tlsc *tls.Config, timeout time.Duration, headers http.Header) (net.Conn, error) {
			return HahaConnect(uri.Host, tlsc, timeout)
		},
	)

      BrokerURL := "haha://127.0.0.1:5883"
      
}

### DIFF
--- a/customdialer.go
+++ b/customdialer.go
@@ -1,0 +1,60 @@
+package mqtt
+
+import (
+	"crypto/tls"
+	"errors"
+	"net"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+)
+
+type CustomDialer = func(uri *url.URL, tlsc *tls.Config, timeout time.Duration, headers http.Header) (net.Conn, error)
+
+type CustomDialerMgr struct {
+	lock    sync.Mutex
+	dialers map[string]CustomDialer
+}
+
+var (
+	dialerMgr *CustomDialerMgr
+)
+
+func init() {
+	dialerMgr = &CustomDialerMgr{
+		dialers: make(map[string]CustomDialer),
+	}
+}
+
+func AddCustomDialer(schema string, fn CustomDialer) error {
+	return dialerMgr.AddDialer(schema, fn)
+}
+
+func GetCustomDialer(schema string) CustomDialer {
+	return dialerMgr.GetDialer(schema)
+}
+
+func (t *CustomDialerMgr) AddDialer(schema string, fn CustomDialer) error {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	if _, ok := t.dialers[schema]; ok {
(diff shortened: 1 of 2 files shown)


MattBrittan:
Thanks for the PR. Unfortunately I cannot progress this until you have signed the ECA (see the commit checks and readme for details). 

However I have taken a quick look and am going to suggest a different approach (happy to discuss this!). Please don't take this the wrong way, your approach is fine and would work, but I think the below may be a better fit. 

My aims with this suggestion are:
 * Minimise the amount of code in this repo (more code =  more maintenance/complexity/potential bugs) and this change meets a niche requirement.
 * Consistent user experience (currently all configuration is done in  `ClientOptions` so that is where users are likely to look for this kind of setting).
 * Avoid [`init()` and globals](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html).
 * Maximise flexibility -  adding different schema types meets your goal but I suspect it may not meet the needs of others (e.g. some weird proxy requirement, [port knocking](https://en.wikipedia.org/wiki/Port_knocking) etc).

I'm also going to suggest that we allow `openConnection` to be replaced in its entirety rather than just adding in new URL handler types.

So lets start with a `ClientOption`:

```go
type openConnectionFunc func(uri *url.URL, tlsc *tls.Config, timeout time.Duration, headers http.Header, websocketOptions *WebsocketOptions) (net.Conn, error)
.
.
.
func NewClientOptions() *ClientOptions {
openConnectionFunc: openConnection // default to inbuilt function
.
.
.

// SetOpenConnectionFunc replaces the inbuilt function that establishes a connection with a custom function.
// The passed in function should return an open `net.Conn` or an error (see the existing openConnection function for an example)
func (o *ClientOptions) SetOpenConnectionFunc(fn openConnectionFunc) *ClientOptions {
	o.openConnectionFunc = fn
	return o
}
```

and then we can replace  the call to   
`conn, err = openConnection(broker, tlsCfg, c.options.ConnectTimeout, c.options.HTTPHeaders, c.options.WebsocketOptions)` in `client.go` with:

`c.Options.openConnectionFunc(broker, tlsCfg, c.options.ConnectTimeout, c.options.HTTPHeaders, c.options.WebsocketOptions)`

I think that this would meet your needs along with some other potential future requirements whilst limiting the level of change. Thoughts?

carr123:
i agree with you.  your way is much better.
i will check what is ECA ,and summit a new PR. :)

MattBrittan:
OK thanks - I'll close this pull request.

Issue #459, opened 2020-10-26
Auto-ack of messages in router.go renders API call mqtt.Message.Ack() useless


tonygunter:
I attempted to verify that I could consume a message and pass it to a worker for processing.  If the processing successfully completes, only then do I want to ack the message.  However, I notice that the messages are acked during consumption by router.go in three places inside the matchAndDispatch function.

I believe this is a bug.  If the intended purpose is to ack the message only *after* the message has been completely processed, then these internal acks are bypassing that functionality.  If the intended purpose is to ack the message immediately after consumption but before processing, then exposing mqtt.Message.Ack() serves no purpose and executes no code (because it's inside a once.Do({}) clause).

tonygunter:
This bug is similar to #396, but the description of the problem is different.  I'm consuming my messages then passing them off to worker processes on a go channel, and expecting to be able to Ack() them there.  He's requesting custom consumers or error-based execution of auto-ack.  I suspect that automatic acks by router.go are the defect, not the solution.

MattBrittan:
I think the two issues are essentially the same. Unfortunately the automatic call to `Ack` is required because removing it would silently break existing code (that is what happened when the change was [initially released](https://github.com/eclipse/paho.mqtt.golang/issues/247)). I don't think that adding a return value to `MessageHandler` is an option (would break existing code) so I see the options as follows:

* Add a new option (say `NoAutoAck` which defaults to `false`). It would need to be clear that this 
* Add an interface (say `AutoAcker` with the function `AutoAckOff()`) that `PublishPacket` implements - calling this would deactivate the automatic `ack` for the particular message (a slightly messy approach but avoids adding yet another option).

alsm:
This is an interesting one, because by the spec the original behaviour is correct. Once the client library has successfully received the message it should acknowledge it, the resend mechanism isn't designed as a feature to be used by applications, it's to ensure that messages successfully pass the network.

tonygunter:
@alsm I think you're correct.  Receiver initiates the onward delivery of the application message, discards the packet identifier, and issues PUBCOMP.  I suppose that means mqtt.Message.Ack() should not be exposed to the public interface?

h12w:
Manual Ack is not about correctness but the efficiency of the message processing. Most message processing involves persisting the results into a database or another message queuing system, where high throughput could be only achieved by batches/transactions with a reasonable size. Processing message one by one is correct but very inefficient.

[...]

PR #578, opened 2022-01-06
Support for AutoAckOff

shivamkm07:
This is to support disabling of Automatic Acking of messages by mqtt client. A similar solution has been merged in V5 client(https://github.com/eclipse/paho.golang/commit/a47276e020a1bc234d42eee3fc0cbcf79b140f4e). While V5 solution also takes care of Acks being sent in order, this doesn't seem a necessity as discussed [here](https://github.com/eclipse/paho.mqtt.golang/issues/459#issuecomment-864384213). 

Issue Resolved: #459

### DIFF
--- a/options.go
+++ b/options.go
@@ -104,7 +104,7 @@ type ClientOptions struct {
 	MaxResumePubInFlight    int // // 0 = no limit; otherwise this is the maximum simultaneous messages sent while resuming
 	Dialer                  *net.Dialer
 	CustomOpenConnectionFn  OpenConnectionFunc
-	AutoAck                 bool
+	AutoAckDisabled         bool
 }
 
 // NewClientOptions will create a new ClientClientOptions type with some
@@ -148,7 +148,7 @@ func NewClientOptions() *ClientOptions {
 		WebsocketOptions:        &WebsocketOptions{},
 		Dialer:                  &net.Dialer{Timeout: 30 * time.Second},
 		CustomOpenConnectionFn:  nil,
-		AutoAck:                 true,
+		AutoAckDisabled:         false,
 	}
 	return o
 }
@@ -449,9 +449,9 @@ func (o *ClientOptions) SetCustomOpenConnectionFn(customOpenConnectionFn OpenCon
 	return o
 }
 
-// SetAutoAck enables or disables the Automated Acking of Messages received by the handler.
-//	By default it is set to true. Setting it to false will disable the auto-ack globally.
-func (o *ClientOptions) SetAutoAck(autoAck bool) *ClientOptions {
-	o.AutoAck = autoAck
+// SetAutoAckDisabled enables or disables the Automated Acking of Messages received by the handler.
+//	By default it is set to false. Setting it to true will disable the auto-ack globally.
+func (o *ClientOptions) SetAutoAckDisabled(autoAckDisabled bool) *ClientOptions {
+	o.AutoAckDisabled = autoAckDisabled
 	return o
 }
--- a/options.go
+++ b/options.go
@@ -104,6 +104,7 @@ type ClientOptions struct {
 	MaxResumePubInFlight    int // // 0 = no limit; otherwise this is the maximum simultaneous messages sent while resuming
 	Dialer                  *net.Dialer
 	CustomOpenConnectionFn  OpenConnectionFunc
+	AutoAck                 bool
 }
 
 // NewClientOptions will create a new ClientClientOptions type with some
@@ -147,6 +148,7 @@ func NewClientOptions() *ClientOptions {
 		WebsocketOptions:        &WebsocketOptions{},
 		Dialer:                  &net.Dialer{Timeout: 30 * time.Second},
 		CustomOpenConnectionFn:  nil,
+		AutoAck:                 true,
 	}
 	return o
 }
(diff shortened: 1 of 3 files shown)


MattBrittan:
Thanks for the PR - before I review it can you please sign the ECA (please let me know if you have any issues). Unfortunately I cannot accept a PR without a signed ECA (the aim being to ensure that all contributions are appropriately licensed).

shivamkm07:
> Thanks for the PR - before I review it can you please sign the ECA (please let me know if you have any issues). Unfortunately I cannot accept a PR without a signed ECA (the aim being to ensure that all contributions are appropriately licensed).

Done!

MattBrittan:
Thanks very much for the PR. I have taken a quick look today and list a few concerns below; please note that these may seem fussy but I think its important that we get this right (once live it is very difficult to change the API and I want to minimise future issues caused by users not understanding how this all works).

It's probably possible to address these concerns by documenting the `Message` interface to ensure users are aware of the potential pitfalls. Given we don't want to make this a breaking change I think that documenting the potential issues may all we can do unless an alternative approach is adopted.

My concerns are:

- Multiple handlers may be fired when a message is received; in this case:
   - If `SetOrderMatters(true)` (default) then the `ACK` will be sent if the first handler called does not call `AutoAckOff()` (regardless of the actions of any further handlers).
  - If `SetOrderMatters(false)` (recommended in the docs) then there is a potential race condition (handlers are called in go routines).
- If the `MessageHandler` calls `AutoAckOff()`, and then returns after starting a go routine that eventually calls `ACK()` then this will lead to a deadlock/unexpected results (just need a warning that ACK must not be called after the handler has returned).
- Where a message is not acknowledged the broker will often only resend it when the connection is re-established (the v3 spec does not specify this but it's what Mosquitto does and what the v5 spec states).

Adding this feature is likely to result in longer running handlers so I feel that its important to note:

- Users need to be aware that long running handlers can cause issues (particularly when `SetOrderMatters(true)`). For example the client will not reconnect until all handlers exit (and I suspect that keep alives may not be sent).
- Due to the way this library handles QOS2 messages duplicates are possible (particularly with long running handlers).

Given these concerns I did consider whether we are going about this in the right way; alternatives would include:
- A `ClientOptions` that turns off all automatic ACK's globally. This would very simple to implement but removes the safety net that automatic ACKs provide.
- Provide a way to flag a route such that messages received on it are not automatically acknowledged. The API for this would require some thought.

Note: I'll copy this to the issue because it would be good to get feedback from users on this.

Volcore:
Hi @MattBrittan and @shivamkm07,

thanks for the PR! I've been running into a similar issue, and stumbled over the auto-ack. Coming from other message passing platforms, like Kafka and PubSub, I was very surprised that auto-ack'ing is even a thing (afaicr it's not even an option for Google's official PubSub library), especially since the Message.Ack() is essentially pointless because of that (at least to the end user of the library).

My preferred option would be to have a ClientOption to at least turn it off (but ideally, not even auto-ack to begin with, but I understand this will create compatibility issues). I'm happy to give a PR for this a go if you two are busy.

One crucial piece that MQTT brokers might be missing to make this really useful is some kind of backoff feature to delay re-delivering messages longer and longer if writes fail (so that we aren't spamming all of our server instances constantly with the same message upon failure), but that's beyond the scope of the client library I think.

shivamkm07:
Hi @MattBrittan and @Volcore, Thanks for reviewing the PR.

I agree adding `AutoAckOff` in Message could lead to race issues, and also it would be redundant to do it for every message. Adding it in `ClientOptions` would disable the AutoAck once globally and then users would be able to Ack the messages as per the handler logic. Therefore, I have updated the PR with `AutoAck` as an option in `ClientOptions`. By default, it is set to true, and the user could set it to false for disabling the automated acking. Let me know if this looks good.

options.go
```diff
@@ -104,6 +104,7 @@ type ClientOptions struct {
 	MaxResumePubInFlight    int // // 0 = no limit; otherwise this is the maximum simultaneous messages sent while resuming
 	Dialer                  *net.Dialer
 	CustomOpenConnectionFn  OpenConnectionFunc
+	AutoAck                 bool
```
MattBrittan, line 107: Sorry to be picky but could we please make this something like 'AutoAckDisabled bool`. The rationale for this is that some users do not use `NewClientOptions` so I'd prefer that the default value for the setting (i.e. `false`) matches the previous behaviour. I realise that this does not read quite as while but feel its worthwhile because based on previous issues some users are just doing `var ClientOptions` (I did add a warning suggesting that this is a bad idea but...). The setter function can remain (just invert the assignment). 
Other than that the PR looks good to go!

shivamkm07: Sure @MattBrittan . That looks good. Making the change.

router.go
```diff
@@ -186,7 +186,9 @@ func (r *router) matchAndDispatch(messages <-chan *packets.PublishPacket, order
 						wg.Add(1)
 						go func() {
 							hd(client, m)
-							m.Ack()
+							if client.options.AutoAckDisabled {
```
MattBrittan, line 189: Sorry - I believe this should now be `if !client.options.AutoAckDisabled` (i.e. if AutoAck is not disabled send the ACK). The same applies to the two below). Hopefully this is the last change!

shivamkm07: Done! Thanks for correcting.

MattBrittan:
Thanks very much for implementing this (sorry it took a few attempts!).

shivamkm07:
@MattBrittan Is there any expected timeline at which the next patch release is scheduled? That would be better than using the latest master commit.

MattBrittan:
Done - I had been holding off because of the major changes to status handling but as they have been in @master for a couple of months without any issues being raised I felt it was worth making a release.

shivamkm07:
Thanks @MattBrittan!!!

PR #583, opened 2022-01-24
Add support for wasm by switching to nhooyr/websocket library.

Bluebugs:
This is only working at the moment with go wasm port and it is my main interest. Nhooyr websocket library allow also for simplifying NewWebsocket code.
 
gopherjs doesn't work for some weird interaction with the token logic channel, but it was beyond my goal with this PR and might require people with better knowledge about gopherjs.


MattBrittan:
This is a fairly large change to meet a relatively niche need so I'd appreciate it if you could raise an issue so this change is more visible and can be discussed.

Note that you can accomplish this without changes to the library by using the new `ClientOptions.SetCustomOpenConnectionFn` to use provide your own custom method for establishing a connection.

I don't personally use MQTT over websockets so am somewhat ambivalent re the change. However I do note that `github.com/gorilla/websocket` has been looking for a new maintainer for a couple of years (but issues are still being handled and PR's merged ). I had some initial concerts re the number of dependencies `nhooyr/websocket` introduces but it does look like the vast majority of these are for the tests only.

Adams214565 (approved):

PR #586, opened 2022-02-11
Disconnect refactor

ptsneves:
The first work on refactoring the client towards getting insights on how to improve it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -452,6 +452,8 @@ func (c *client) attemptConnection() (net.Conn, byte, bool, error) {
 // reusing the `client` may lead to panics. If you want to reconnect when the connection drops then use
 // `SetAutoReconnect` and/or `SetConnectRetry`options instead of implementing this yourself.
 func (c *client) Disconnect(quiesce uint) {
+	defer c.disconnect()
+
 	status := atomic.LoadUint32(&c.status)
 	if status == connected {
 		DEBUG.Println(CLI, "disconnecting")
@@ -480,7 +482,6 @@ func (c *client) Disconnect(quiesce uint) {
 		c.setConnected(disconnected)
 	}
 
-	c.disconnect()
 }
 
 // forceDisconnect will end the connection with the mqtt broker immediately (used for tests only)
--- a/client.go
+++ b/client.go
@@ -455,9 +455,9 @@ func (c *client) Disconnect(quiesce uint) {
 	defer c.disconnect()
 
 	status := atomic.LoadUint32(&c.status)
+	c.setConnected(disconnected)
 	if status == connected {
 		DEBUG.Println(CLI, "disconnecting")
-		c.setConnected(disconnected)
 
 		dm := packets.NewControlPacket(packets.Disconnect).(*packets.DisconnectPacket)
 		dt := newToken(packets.Disconnect)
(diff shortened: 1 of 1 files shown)


MattBrittan:
Thanks - however I will need you to sign the ECN before I'm able to accept your PR's (see the [readme](https://github.com/eclipse/paho.mqtt.golang#contributing) for more details).

ptsneves:
Oki i signed the ECA but it still says it is not covered by the ECA. Does it take some time to be active?

client.go
```diff
@@ -452,35 +452,29 @@ func (c *client) attemptConnection() (net.Conn, byte, bool, error) {
...
-		select {
-		case c.oboundP <- &PacketAndToken{p: dm, t: dt}:
-			disconnectSent = true
-		case <-c.commsStopped:
-			WARN.Println("Disconnect packet could not be sent because comms stopped")
-		case <-time.After(time.Duration(quiesce) * time.Millisecond):
-			WARN.Println("Disconnect packet not sent due to timeout")
-		}
+	defer c.disconnect()
+	c.setConnected(disconnected)
```
MattBrittan, line 456: This will set c.status to `disconnected` meaning that `if status != connected {` will always be true...

MattBrittan (changes requested):
Changes are needed before this can be accepted.

client.go
ptsneves: Yep, but unless I am missing something subtle, Isnt this already the case in the current code?
```
if status == connected {
 c.setConnected(disconnected)
....
}
else {
   // Irrelevant print
   c.setConnected(disconnected)
}
```

MattBrittan: >unless I am missing something subtle

It's not really subtle. The current code is:

```
status := atomic.LoadUint32(&c.status)
if status == connected {
	DEBUG.Println(CLI, "disconnecting")
	c.setConnected(disconnected)
        // Send DISCONNECT etc...
} else {
	WARN.Println(CLI, "Disconnect() called but not connected (disconnected/reconnecting)")
	c.setConnected(disconnected)
}
```

So this checks the status and IF the client is connected it sends a DISCONNECT packet (calling `c.setConnected(disconnected)` immediately **after** the check regardless of whether the client is connected or not).

Your change is:

```
c.setConnected(disconnected)
status := atomic.LoadUint32(&c.status)
if status != connected {
    // Do stuff...
}
```

This calls `c.setConnected(disconnected)` **BEFORE** checking the status so what you are effectively doing is setting `status` to `disconnected` and then, after changing the value, checking if `status != connected` (this will **always** be true because you just set it to `disconnected`). This can be summarised as:

```
c.status = disconnected
status := c.status
if status != connected {
    // As you have set status to disconnected this branch will ALWAYS be taken and the client will never disconnect cleanly
   return
}
// Unreachable code....
```

You could fix this by changing the order of your statements:

```
status := atomic.LoadUint32(&c.status)
c.setConnected(disconnected)
if status != connected {
}
```

However I'm not altogether convinced that this is easier to read (it does make the race condition more obvious but, as I mentioned in the issue, the way `c.status` is handled is far from ideal (the result of many changes over the years by a large number of contributors).

MattBrittan:
Thanks for your contribution.

PR #588, opened 2022-03-15
Avoid race in Subscribe Token result

ptsneves:
maps are passed/returned by internal pointer instead of value
so there can be concurrent use of the resource resulting
in a race. Fixed and test case included.


ptsneves:
Hey Matt,
Thank you for your thoughtful comment. I can see your point of view, although given the reproducible issues with concurrency in paho I am taking a fix as I see it approach. So it is a hypothetical issue backed by a reproducible test. The goal is to get the correctness right, add the necessary tests, and then start doing more invasive changes.

I will close the pull request and leave it on my local repos. I will probably get back to you with more proposals if you do not mind. Although now I have a better feel on what is the philosophy of the project and I should not cause much hassle :)

PR #615, opened 2022-10-06
Update main.go

Gsolisen:
add reference mqtt name on import golang package

### DIFF
--- a/cmd/simple/main.go
+++ b/cmd/simple/main.go
@@ -24,7 +24,7 @@ import (
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


MattBrittan:
This seems unnecessary? (code compiles fine without the extra label).
Note that, in any event, I cannot accept a PR until you have completed the ECA process (see check above).

MattBrittan:
I'm going to close this PR because I'd need an explanation as to why its needed and the ECA signed before it can go anywhere.

PR #624, opened 2022-12-22
Allow MQTT username/password in URL when connecting via Websockets

### LINKED ISSUE #623: Websocket connections with password in URL fail
This library allows the MQTT user/password to be specified in the URL ie. `wss://user:password@mqtt.broker.com:443/mqtt`. Unfortunately this does not work with web-socket connections because the websocket library [rejects url's containing a username/password](https://github.com/gorilla/websocket/blob/master/client.go#L184). 

This is annoying because including the user/password in the URL could be a useful way to pass the MQTT username/password...

MattBrittan:
Clear URL User before creating websocket connection (allowing MQTT username/password to be specified in URL). If u.User is not nil then the Gorilla Websocket library will reject the connection attempt.

Closes #623

### DIFF
--- a/netconn.go
+++ b/netconn.go
@@ -40,10 +40,14 @@ import (
 func openConnection(uri *url.URL, tlsc *tls.Config, timeout time.Duration, headers http.Header, websocketOptions *WebsocketOptions, dialer *net.Dialer) (net.Conn, error) {
 	switch uri.Scheme {
 	case "ws":
-		conn, err := NewWebsocket(uri.String(), nil, timeout, headers, websocketOptions)
+		dialURI := uri // #623 - Gorilla Websockets does not accept URL's where uri.User != nil
+		uri.User = nil
+		conn, err := NewWebsocket(dialURI.String(), nil, timeout, headers, websocketOptions)
 		return conn, err
 	case "wss":
-		conn, err := NewWebsocket(uri.String(), tlsc, timeout, headers, websocketOptions)
+		dialURI := uri // #623 - Gorilla Websockets does not accept URL's where uri.User != nil
+		uri.User = nil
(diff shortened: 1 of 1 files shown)


netconn.go
```diff
@@ -40,10 +40,14 @@ import (
 func openConnection(uri *url.URL, tlsc *tls.Config, timeout time.Duration, headers http.Header, websocketOptions *WebsocketOptions, dialer *net.Dialer) (net.Conn, error) {
 	switch uri.Scheme {
 	case "ws":
-		conn, err := NewWebsocket(uri.String(), nil, timeout, headers, websocketOptions)
+		dialURI := uri // #623 - Gorilla Websockets does not accept URL's where uri.User != nil
+		uri.User = nil
+		conn, err := NewWebsocket(dialURI.String(), nil, timeout, headers, websocketOptions)
```
Tieske, line 45: Unless I'm mistaken, this code alters the original url (especially since it was passed in as a pointer to this function).

Shouldn't it be like this:
```go
		dialURI := *uri // #623 - Gorilla Websockets does not accept URL's where uri.User != nil
		dialURI.User = nil
		conn, err := NewWebsocket(dialURI.String(), nil, timeout, headers, websocketOptions)
```

If so, then the same issue is on the `"wss"` branch below

Tieske: ping @MattBrittan (apologies if inappropriate, but doing this since I'm responding to a closed PR)

MattBrittan: Well spotted; this change did work (This was a quick fix, made on a customers site, that I needed to move to websockets urgently to bypass a firewall :-) ) but I can now see that it would have failed on subsequent connections (funnily enough I did have an issue with that site but put it down to websockets :-) ). I'll push a patch through shortly.

Issue #589, opened 2022-03-17
Add Option for reconnect Back-Off (to prevent continual reconnection attempts when connection closed by remote host)


chess555:
Due to the way connect/reconnects are handled, no delay is used when a remote host closes a connection. 

The simplest example is publishing using a bad topic with QoS > 0
  The client will store the bad message, and get stuck in a cycle of connect, publish, connection closed, reconnect, ...
  
Similar behavior has been noted when:
  Connection is dropped due to duplicate clientids in a broker
  Performing an operation (subscribe/publish) on an AWS data broker when your device lacks the policy settings, or has had policy permissions revoked.

MattBrittan:
I'm not sure that there is a solution to this unfortunately; as per [the spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc384800415):

>If a Server implementation does not authorize a PUBLISH to be performed by a Client; it has no way of informing that Client. It MUST either make a positive acknowledgement, according to the normal QoS rules, or close the Network Connection [MQTT-3.3.5-2].

So most 'well behaved' brokers will complete the handshake and throw the message away. While disconnecting is given as an option the client has no way to determine why the broker dropped the connection and, as such, cannot remove the offending message from it's store (the spec does not make any provision for 'rejected' messages).

The same applies in other situations (e.g. another client connects with the same client ID); the broker just drops the connection and we have no way to differentiate that from a network issue.

The above means that I don't think we really have any option other than to attempt to reconnect immediately and send through any queued messages. If you can suggest an alternative I'll definitely consider it (but most libraries use the same approach).

chess555:
Sorry for not being clear, but the only issue I'm pointing at is the reconnect frequency performed when a close event occurs. (The examples were just meant as a situation where this occurs)

When a remote host closes a connection in this manner, I'm seeing upwards of 50-100 connect/disconnect events per second, and am concerned about this behavior (particularly on constrained networks)

In my case, I was able to use the Reconnect handler to add a delay after receiving an an io.EOF event.

MattBrittan:
No worries - I have amended the title so that it more clearly covers what I believe you are requesting.

The initial reconnection attempt needs to be immediate (because the issue may be a momentary network glitch) but I agree that continually attempting to reconnect is counter productive. Some form of Back-Off algorithm (reset after the connection has been up for more than a user specified time) would be beneficial.

[...]

PR #625, opened 2022-12-26
Add back-off controller for sleep time of reconnection  when connection lost is detected immediately after connecting. #589

tomatod:
MaxConnectRetryInterval is added to ClientOptions related to #589.

The sleep time is increased progressively between connection attempts up to MaxConnectRetryInterval.
If ConnectRetryInterval >= MaxConnectRetryInterval, ConnectRetryInterval is kept between the attempts.

### DIFF
--- a/backoff.go
+++ b/backoff.go
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2021 IBM Corp and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    https://www.eclipse.org/legal/epl-2.0/
+ * and the Eclipse Distribution License is available at
+ *   http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ *    Matt Brittan
+ *    Daichi Tomaru
+ */
+
+package mqtt
+
+import (
+	"sync"
+	"time"
+)
+
+// Controller for sleep with backoff when the client attempts reconnection
+// It has statuses for each situations cause reconnection.
+type backoffController struct {
+	sync.RWMutex
+	statusMap map[string]*backoffStatus
+}
+
+type backoffStatus struct {
+	lastSleepPeriod time.Duration
+	lastErrorTime   time.Time
+}
+
+func newBackoffController() *backoffController {
+	return &backoffController{
+		statusMap: map[string]*backoffStatus{},
+	}
+}
+
+// Calculate next sleep period from the specified parameters.
+// Returned values are next sleep period and whether the error situation is continual.
+// If connection errors continuouslly occurs, its sleep period is exponentially increased.
+// Also if there is a lot of time between last and this error, sleep period is initialized.
+func (b *backoffController) getBackoffSleepTime(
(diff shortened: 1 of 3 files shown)


MattBrittan:
The PR looks useful but does not appear to address the related issue. You PR allows provides more flexibility re the delay between unsuccessful connections but the issue related to where the connection is successful but is dropped soon after being made due to an invalid publish, another client with the same ID etc. Do you want to try and address the larger issue or just go with this smaller change?

tomatod:
MattBrittan, Thanks for your comment. I wasn't enough to check issue. I'll try to reproduce other situation and work on that again. If it's too difficult for me, I may push smaller modification.

tomatod:
@MattBrittan 
I fixed PR, Could you review it again?

### Summary
- I worked on the bellow No.3 point not No.1, considering the order of implementation. When I work in No.1, I'll make new PR.
https://github.com/eclipse/paho.mqtt.golang/issues/589#issuecomment-1366294006

- I added new type backoffController for handling sleep time by some situations universally. 

### Implementation point
backoffController type has mutex in view of thread safe. I thought reconnect method may be called from multi threads. 

According to the implementation below, internalConnLost method is expected to be stopped (return), when it's called by multi thread. But both the errors errConnLossWhileDisconnecting and errAlreadyHandlingConnectionLoss seem to be not returned anywhere although they are defined. So "c.status.ConnectionLost" (L515) may not return any error and continue if that is call by secondary thread.
https://github.com/eclipse/paho.mqtt.golang/blob/master/client.go#L515-L522

Note that internalConnLost method may be called in the bellow.
https://github.com/tomatod/paho.mqtt.golang/blob/master/ping.go#L73

I was wondering where to embed directly a thread-safe store (in types client or connectionStatus ?), but there seems to be no appropriate place. So I added that as new type.

MattBrittan:
Sorry about the delay in getting to this. Your PR looks OK to me (and all tests pass) so I'm going to accept it (it's a bit more complicated than I'd expected but I can see why you have made the decisions you did).

>reconnect method may be called from multi threads.
It should not be (the status code should prevent this but it's from the most recent change so it's possible I've missed something). However a mutex is worthwhile , the code seems pretty generic so may be useful elsewhere so it's better to make it thread safe (and its performance is not really of critical importance)..

>But both the errors errConnLossWhileDisconnecting and errAlreadyHandlingConnectionLoss seem to be not returned anywhere although they are defined.

You are correct - however this is not a big deal because the only real impact is that an extra error is logged (assuming logging is enabled). The check for these errors was really just to prevent logging of thinbgs that are expected to occur as part of normal operations. I believe the situations I was trying to stop logging for are [here](https://github.com/eclipse/paho.mqtt.golang/blob/d174b9ad5a9a0463dc8013047d6b0a9114d59e2f/status.go#L198) and [here](https://github.com/eclipse/paho.mqtt.golang/blob/d174b9ad5a9a0463dc8013047d6b0a9114d59e2f/status.go#L242).

fsudaman:
Hi @MattBrittan any plan for next release soon that includes this PR? Thanks

MattBrittan:
@fsudaman - sorry things have been pretty stable so I had not considered a release. It's been eight months so may as well! - Done

fsudaman:
@MattBrittan we had some issues with huge amount of retries causing a bit of congestion at scale so this really help. Thank you for the quick response and much appreciated.

PR #646, opened 2023-08-09
Wrap connection network errors

adriansmares:
This PR implements [error wrapping](https://go.dev/blog/go1.13-errors) for the connection network errors, in order to allow them to be asserted using the [errors](https://pkg.go.dev/errors) package, specifically [errors.Is](https://pkg.go.dev/errors#Is).

Pretty much all of the boiler plate proposed in this PR can be handled using Go 1.20 [multiple error wrapping](https://tip.golang.org/doc/go1.20#errors), but since that's a 1.20+ feature, we manually handle this instead. When Go 1.20 becomes the minimum supported version, most of this PR can be reduced to `fmt.Errorf("%w : %w", packets.ConnErrors[rc], err)`.

Some formatting is changed due to the fact that I automatically format files using [gofumpt](https://github.com/mvdan/gofumpt) - please let me know if you would like for these changes to be left out.

### DIFF
--- a/client.go
+++ b/client.go
@@ -142,7 +142,7 @@ type client struct {
 	workers      sync.WaitGroup // used to wait for workers to complete (ping, keepalive, errwatch, resume)
 	commsStopped chan struct{}  // closed when the comms routines have stopped (kept running until after workers have closed to avoid deadlocks)
 
-	backoff      *backoffController
+	backoff *backoffController
 }
 
 // NewClient will create an MQTT v3.1.1 client with all of the options specified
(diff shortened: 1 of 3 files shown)


MattBrittan:
Was concerned that this is a breaking change because `ErrorNetworkError` is public. However reviewing the code we never return an error of that type due to the check:

```
// Maintain same error format as used previously
if rc != packets.ErrNetworkError { // mqtt error
	err = packets.ConnErrors[rc]
} else { // network error (if this occurred in ConnectMQTT then err will be nil)
	err = fmt.Errorf("%s : %s", packets.ConnErrors[rc], err)
}
```

As a result I think we are OK with this.

However the change does add quite a bit of code which, as you say, is not required if we drop support for earlier versions of Go. With the release of 1.21 I don't have a problem with following the Go team policy and only supporting 1.20 and 1.21. As such might I suggest that we go with the simpler version and Go 1.20? (I would accept into `master` but not push out a new release for a month or so to provide time for people to migrate off 1.18/1.19; I think most go devs update relatively quickly).

adriansmares:
> As such might I suggest that we go with the simpler version and Go 1.20?

I have converted this PR to the Go 1.20 multi wrap format, and updated the minimum Go version to 1.20, along with the dependencies.

PR #665, opened 2024-01-07
fix: deprecation warnings for ioutil

vruge:


MattBrittan:
Sorry - I realise this is only a small change, but I cannot accept any changes without a signed ECA (see check above). This is to ensure that the code is correctly licensed.

vruge:
> Sorry - I realise this is only a small change, but I cannot accept any changes without a signed ECA (see check above). This is to ensure that the code is correctly licensed.

done

MattBrittan:
Note that this introduced an issue - see #673 (I should have tested before accepting the PR but assumed you would have compiled the code).

PR #671, opened 2024-03-21
Replace the time.After with the timer for efficiency.

DVasselli:
The old `time.After` was not stopping the internal timer creating a huge resource overhead for frequently publishing clients. This way we stop the timer in the `defer` increasing efficiency.

Signed-off-by: Daniele Vasselli <vasselli.daniele@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -799,12 +799,10 @@ func (c *client) Publish(topic string, qos byte, retained bool, payload interfac
 		if publishWaitTimeout == 0 {
 			publishWaitTimeout = time.Second * 30
 		}
+
 		t := time.NewTimer(publishWaitTimeout)
-		defer func() {
-			if !t.Stop() {
-				<-t.C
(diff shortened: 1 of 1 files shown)


MattBrittan:
Thanks for the change; however this will deadlock if the timer does fire [example](https://go.dev/play/p/v6yvA5tRYn4). As the timer is not reused there is no real need to drain the channel.

DVasselli:
Ohhh I totally misunderstood the documentation of `t.Stop()`. I pushed it without the drain nd we'll let the GC take care of the channel after return

MattBrittan:
You are [not the first](https://github.com/golang/go/issues/27169) person to do so. Thanks for the PR.

DVasselli:
Happy to have learned something new! Any chance you can make a new tag for this so I can import it?

PR #672, opened 2024-03-31
remove unused error errAlreadyHandlingConnectionLoss

wcsiu:
# Chanages
* Remove unused errAlreadyHandlingConnectionLoss.

### DIFF
--- a/client.go
+++ b/client.go
@@ -516,7 +516,7 @@ func (c *client) internalConnLost(whyConnLost error) {
 	DEBUG.Println(CLI, "internalConnLost called")
 	disDone, err := c.status.ConnectionLost(c.options.AutoReconnect && c.status.ConnectionStatus() > connecting)
 	if err != nil {
-		if err == errConnLossWhileDisconnecting || err == errAlreadyHandlingConnectionLoss {
+		if err == errConnLossWhileDisconnecting {
 			return // Loss of connection is expected or already being handled
 		}
 		ERROR.Println(CLI, fmt.Sprintf("internalConnLost unexpected status: %s", err.Error()))
--- a/status.go
+++ b/status.go
@@ -67,9 +67,11 @@ func (s status) String() string {
 	}
 }
 
-type connCompletedFn func(success bool) error
-type disconnectCompletedFn func()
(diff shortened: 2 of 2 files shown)


MattBrittan:
Sorry - I will be unable to accept this without a signed ECA (see the readme and the above link for more info). 

The important bit of this PR would be `|| c.status == disconnected`; however `connectionStatus` was designed such that transitioning from `connecting` to `disconnected` should not happen (if it does then there are potential race conditions). Consider the situation where the user calls `Disconnect` and then `Connect` again before `connected` gets called (seems possible from a quick review of the code).

However I believe that `connectionStatus` already accounts for this. `Disconnecting()` waits on `c.actionCompleted` before completing if the status was `connecting` or `reconnecting` (which should prevent the status from becoming `disconnected` before `connected` is called. 

Now it is possible that I'm missing something (wrote this code a while back, it's quite complex, and I've just scanned it). Are you able to replicate the issue you are attempting to address?

wcsiu:
> Sorry - I will be unable to accept this without a signed ECA (see the readme and the above link for more info).
> 
> The important bit of this PR would be `|| c.status == disconnected`; however `connectionStatus` was designed such that transitioning from `connecting` to `disconnected` should not happen (if it does then there are potential race conditions). Consider the situation where the user calls `Disconnect` and then `Connect` again before `connected` gets called (seems possible from a quick review of the code).
> 
> However I believe that `connectionStatus` already accounts for this. `Disconnecting()` waits on `c.actionCompleted` before completing if the status was `connecting` or `reconnecting` (which should prevent the status from becoming `disconnected` before `connected` is called.
> 
> Now it is possible that I'm missing something (wrote this code a while back, it's quite complex, and I've just scanned it). Are you able to replicate the issue you are attempting to address?

I think you are right about it that it will never reach disconnected state for that conditional check. My mistake on that.

MattBrittan:
Closing this as I cannot accept it without an ECA (see [here](https://api.eclipse.org/git/eca/status/gh/eclipse/paho.mqtt.golang/672) for more details). Realise it's only a very small change but the rules are the rules (and are there to protect the integrity of the project). If you are able to sort out the ECA then this can be reopened.

Issue #675, opened 2024-05-08
goroutine leak when connectionUp(true) return error


kiqi007:
**Issue Description**: Encountered a goroutine leak scenario when using the MQTT package.
```go
	for {
		select {
		case <-m.checkDone:
			hlog.Info("mqtt check done")
			return
		default:
			if m.client == nil || !m.client.IsConnectionOpen() {
				if nil != m.client {
					m.client.Disconnect(1000)
				}
				err := m.Connect() // create a new client, and do connect
				if nil != err {
					hlog.Error("mqtt Connect error", elog.String("host", m.host), elog.Int("port", m.port), elog.FieldErr(err))
				} else {
					hlog.Info("mqtt Connect success", elog.String("host", m.host), elog.Int("port", m.port))
				}
			}
			<-m.connCheckTicker.C
		}
	}
``` 

``` 
gorouine pprof:
goroutine profile: total 2787
1339 @ 0xe2e556 0xe3edac 0xe3ed86 0xe57ae5 0xe77032 0x16e6c25 0x16e43dc 0xe5ca61
#	0xe57ae4	sync.runtime_Semacquire+0x24							/usr/local/go/src/runtime/sema.go:56
#	0xe77031	sync.(*WaitGroup).Wait+0x51							/usr/local/go/src/sync/waitgroup.go:136
#	0x16e6c24	github.com/eclipse/paho%2emqtt%2egolang.(*client).startCommsWorkers+0x4a4	/home/jenkins/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.4.2/client.go:608
#	0x16e43db	github.com/eclipse/paho%2emqtt%2egolang.(*client).Connect.func1+0x49b		/home/jenkins/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.4.2/client.go:282

1339 @ 0xe2e556 0xe3edac 0xe3ed86 0xe57c05 0xe755a5 0x16e7cc9 0x16e7c9e 0x16e5cc5 0x16e5c06 0x16e5b87 0xe5ca61
#	0xe57c04	sync.runtime_SemacquireMutex+0x24						/usr/local/go/src/runtime/sema.go:71
#	0xe755a4	sync.(*Mutex).lockSlow+0x164							/usr/local/go/src/sync/mutex.go:162
#	0x16e7cc8	sync.(*Mutex).Lock+0xa8								/usr/local/go/src/sync/mutex.go:81
[...]

PR #678, opened 2024-05-19
fix: issue 675：goroutine leak when connectionUp(true) return error

kiqi007:
bug fix: https://github.com/eclipse/paho.mqtt.golang/issues/675
- Skip the check for Disconnect when the connection is being established.
- Treat this scenario (Disconnect during connection) as if the Disconnect occurred after connectionUp(true) completed.

### DIFF
--- a/client.go
+++ b/client.go
@@ -604,14 +604,25 @@ func (c *client) startCommsWorkers(conn net.Conn, connectionUp connCompletedFn, 
 	// The connection is now ready for use (we spin up a few go routines below). It is possible that
 	// Disconnect has been called in the interim...
 	if err := connectionUp(true); err != nil {
-		DEBUG.Println(CLI, err)
-		close(c.stop) // Tidy up anything we have already started
-		close(incomingPubChan)
-		c.workers.Wait()
-		c.conn.Close()
-		c.conn = nil
-		return false
+		ERROR.Println(CLI, err)
+
+		/* issue 675：goroutine leak when connectionUp(true) return error
+		 *   Only when status == disconnecting will this logic be executed.
+		 * The goroutine that changes the status to disconnecting should be
+		 * responsible for resource cleanup (which is indeed how it is done).
+		 *
+		 *   Being disconnected right when the connection is established is a special case.
+		 * Even if we remove this check for connectionUp(true), the program must still function correctly,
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -604,14 +604,25 @@ func (c *client) startCommsWorkers(conn net.Conn, connectionUp connCompletedFn,
...
+		 *   Only when status == disconnecting will this logic be executed.
+		 * The goroutine that changes the status to disconnecting should be
+		 * responsible for resource cleanup (which is indeed how it is done).
+		 *
+		 *   Being disconnected right when the connection is established is a special case.
+		 * Even if we remove this check for connectionUp(true), the program must still function correctly,
+		 * as if a Disconnect event occurred immediately after connectionUp(true) completed.
+		 */
+
+		//close(c.stop) // Tidy up anything we have already started
```
MattBrittan, line 619: So, in summary, this PR will change things so that:

 if a connection is in progress when `Disconnect` is called then the connection attempt will be completed (including calling `OnConnect` and spinning up goroutines) and then the `Disconnect` operation will proceed after this (so, effectively a normal disconnect). In this case `connectionStatus.Disconnecting()` will wait until the connection process has completed before returning (and allowing the disconnection process to proceed). `startCommsWorkers` maintains a lock on `connMu` throughout the process, so `stopCommsWorkers` will be blocked until it completes.

My feeling is that this should work (but it's quite difficult to trace all possibilities). However I'd prefer to see the old code removed rather than being commented out (it will be available in the git history) and would like to see a comment something like the above (i.e. we will allow the connection to complete before the Disconnect is allowed to proceed).

I think I'm OK to accept this but if you are able to add something like the following I'd appreciate it. Adding a test for this would be absolutely fantastic (but I'm thinking it might be quite difficult to write).

kiqi007: Of course, I have followed your suggestions. 
For testing, I simulated this scenario locally using the method I previously used to debug the issue, and it works.

MattBrittan: Thanks.

>For testing, I simulated this scenario locally using the method I previously used to debug the issue, and it works.

This comment was more about adding a test (that does not pass with the old code but does with the new).  The main purpose of this is to help prevent regressions (the state management stuff is pretty complex and Disconnect is not used all that often so it would be east to reintroduce the issue). Anyway I'll accept this PR now and if you do have a test that could be added later.

PR #680, opened 2024-07-18
Add js websocket supported

LittleKey:
Supported compile to js&&wasm:
`GOOS=js GOARCH=wasm go build xxx.go`

And already test passed on browser.


LittleKey:
Duplication with: https://github.com/eclipse/paho.mqtt.golang/pull/583

PR #682, opened 2024-07-22
fix(ping outstanding and ping request check): check the ping timeout …

CaiYueTing:
check the ping timeout after the check keepalive

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -66,12 +66,11 @@ func keepalive(c *client, conn io.Writer) {
 					}
 					c.lastSent.Store(time.Now())
 					pingSent = time.Now()
+				} else if time.Since(pingSent) >= c.options.PingTimeout {
+					CRITICAL.Println(PNG, "pingresp not received, disconnecting")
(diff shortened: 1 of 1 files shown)


MattBrittan:
Can you please explain what issue this solves? From a quick look this change would mean that the ping timeout would only be fired when `time.Since(lastSent) >= time.Duration(c.options.KeepAlive*int64(time.Second))` AND `time.Since(pingSent) >= c.options.PingTimeout` wherease the current functionality only requires the second condition to be true (by default `PingTimeout` is 10s and `options.KeepAlive` is 30s).

CaiYueTing:
Hi @MattBrittan,
I'm thinking, if data's always transmitted, why do we still need to use ping requests?

Before sending a ping, we should check the time of the last update. If no data has been transmitted within the keepalive interval, then we can send a ping request.

If I have any misunderstandings, please correct me.

MattBrittan:
The code already does that, you have just moved the code that handles the disconnection when no reply is received. Note that I'm not going to do anything further with this without a more detailed explanation and signed ECA (could not accept the PR without the ECA in any event).

Issue #690, opened 2024-09-03
Dockerfile not building image because of go.sum not updated with dependencies


JefJrFigueiredo:
In this directory https://github.com/eclipse/paho.mqtt.golang/tree/master/cmd/docker , 
I've followed the instructions, which said: 
"Ensure that you have docker and docker-compose installed.

To start everything up change into the cmd/docker folder and run:

`docker-compose up --build --detach` "

But I've got this error:

=> ERROR [sub build-env 4/4] RUN go build -gcflags "all=-N -l" -o /sub                                                                                                                               0.8s
 => ERROR [pub build-env 4/4] RUN go build -gcflags "all=-N -l" -o /pub                                                                                                                               0.7s
------
 > [sub build-env 4/4] RUN go build -gcflags "all=-N -l" -o /sub:
0.695 main.go:27:2: missing go.sum entry for module providing package github.com/eclipse/paho.mqtt.golang (imported by subscriber); to add:
0.695   go get subscriber
------
------
 > [pub build-env 4/4] RUN go build -gcflags "all=-N -l" -o /pub:
0.696 main.go:26:2: missing go.sum entry for module providing package github.com/eclipse/paho.mqtt.golang (imported by publisher); to add:
0.696   go get publisher
------
failed to solve: process "/bin/sh -c go build -gcflags \"all=-N -l\" -o /pub" did not complete successfully: exit code: 1

I've already solved that by adding `go mod tidy &&` before `go build -gcflags` in cmd/docker/publisher/dockerfile and cmd/docker/subscriber/dockerfile. After that, `docker-compose up --build --detach` runs smoothly:

Attaching to mosquitto-1, pub-1, sub-1
mosquitto-1  | chown: /mosquitto/config/mosquitto.conf: Read-only file system
mosquitto-1  | chown: /mosquitto/config: Read-only file system
mosquitto-1  | chown: /mosquitto/config: Read-only file system
mosquitto-1  | 2024-09-03T08:18:26: mosquitto version 2.0.18 starting
mosquitto-1  | 2024-09-03T08:18:26: Config loaded from /mosquitto/config/mosquitto.conf.
mosquitto-1  | 2024-09-03T08:18:26: Opening ipv4 listen socket on port 1883.
mosquitto-1  | 2024-09-03T08:18:26: Opening ipv6 listen socket on port 1883.
mosquitto-1  | 2024-09-03T08:18:26: mosquitto version 2.0.18 running
[...]

PR #691, opened 2024-09-03
fix: issue 690: updating go dependencies from pub and sub into the containers before building

JefJrFigueiredo:
bug fix: #690 

- By running `go mod tidy &&` before `go build -gcflags` in the dockerfile of pub and sub, the containers are able to be builded and run with the `docker-compose up --build --detach` command.


MattBrittan:
Thanks very much.

Issue #703, opened 2025-03-05
Data race while reconnecting


NiklasCi:
One of our test fails with a data race if we test our logic we implemented if the client reconnects.
It seems that there is a DATA RACE in the client itself. At least the trace does not show any of our go files/packages.

Trace:
```
WARNING: DATA RACE
Write at 0x00c00047e441 by goroutine 147:
  github.com/eclipse/paho%2emqtt%2egolang.(*client).resume()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/client.go:1090 +0x13de
  github.com/eclipse/paho%2emqtt%2egolang.(*client).reconnect()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/client.go:341 +0x973
  github.com/eclipse/paho%2emqtt%2egolang.(*client).internalConnLost.func1.gowrap1()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/client.go:562 +0x44

Previous read at 0x00c00047e441 by goroutine 167:
  github.com/eclipse/paho.mqtt.golang/packets.(*FixedHeader).pack()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/packets/packets.go:267 +0x71
  github.com/eclipse/paho.mqtt.golang/packets.(*PublishPacket).Write()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/packets/publish.go:47 +0x2dd
  github.com/eclipse/paho%2emqtt%2egolang.startOutgoingComms.func1()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/net.go:293 +0xd5e

Goroutine 147 (running) created at:
  github.com/eclipse/paho%2emqtt%2egolang.(*client).internalConnLost.func1()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/client.go:562 +0x60a

Goroutine 167 (running) created at:
  github.com/eclipse/paho%2emqtt%2egolang.startOutgoingComms()
      /home/niklas/go/pkg/mod/github.com/eclipse/paho.mqtt.golang@v1.5.0/net.go:264 +0x3a4
[...]

PR #708, opened 2025-03-15
Return a copy from the memstore to avoid data races

Nocccer:
# Description

To avoid data races that occured in #703, a copy is now returned by the memstore instead a pointer to the packet.

- [x] Extend the `ControlPacket` interface with `Copy() ControlPacket` function.
- [x] Add tests to check if the copy does not share any pointer with the original packet.
- [x] Use the `Copy()` function in the memstore to return a copy.
- [ ] Test changes against the test from #703 that triggered the race condition.

**Note:** This is my private github account, so don't get confused.

### DIFF
--- a/memstore.go
+++ b/memstore.go
@@ -65,7 +65,7 @@ func (store *MemoryStore) Put(key string, message packets.ControlPacket) {
 }
 
 // Get takes a key and looks in the store for a matching Message
-// returning either the Message pointer or nil.
+// returning either a copy of the Message as packets.ControlPacket or nil.
 func (store *MemoryStore) Get(key string) packets.ControlPacket {
 	store.RLock()
 	defer store.RUnlock()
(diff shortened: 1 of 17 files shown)


MattBrittan:
Please note that I cannot accept this without a signed ECA (see the above check for details).

MattBrittan:
Sorry for the delay in looking at this (not a lot of spare time currently). As this changes `ControlPacket` it's technically a breaking change so I want to take a bit more time to consider it (think it's unlikely that anyone relies on this but the ways people use this library frequently surprises me!).

MattBrittan:
@copilot resolve the merge conflicts in this pull request

MattBrittan:
Sorry for the huge delay in looking into this again. I'm going to put a bit more time into looking at alternatives (e.g. adding a `clonePacket` helper in the memory store) because I'm reluctant to make a potentially breaking change. I'm also reluctant to add the extra copy operation into the hot path (may not seem like much but I know there are some performance sensitive users).
Currently thinking about adding comments to make it clear that the packets returned from the store should not be modified, and making a copy before the existing modification (this may be cleaned, and reduces the overall impact).

PR #710, opened 2025-04-09
fix: reduce slice allocations in route dispatch 

alespour:
This PR makes `handlers` slice reused in `router.matchAndDispatch` in order to reduce slice (re)allocations. I think it is safe to reuse the slice since it is only used when `order = true`. Therefore, also added explicit `if order` check for iterating over handlers, although it is safe to keep iterating over nil slice when `order = false`, of course.

It is only a subtle optimization, but appeared when I profiled high-load scenario.

Benchmark with a _single_ route:

before:
```
go test -v -run Benchmark_MatchAndDispatch -bench=. -benchmem -benchtime=10s

Benchmark_MatchAndDispatch-4   	13364338	       884.3 ns/op	     136 B/op	       3 allocs/op
```

after:
```
go test -v -run Benchmark_MatchAndDispatch -bench=. -benchmem -benchtime=10s

Benchmark_MatchAndDispatch-4   	14650513	       808.3 ns/op	     128 B/op	       2 allocs/op
```

I suppose with more routes, without reusing the slice, append would cause more re-allocations.

### DIFF
--- a/router.go
+++ b/router.go
@@ -171,12 +171,12 @@ func (r *router) matchAndDispatch(messages <-chan *packets.PublishPacket, order 
 	}
 
 	go func() { // Main go routine handling inbound messages
+		var handlers []MessageHandler
 		for message := range messages {
 			// DEBUG.Println(ROU, "matchAndDispatch received message")
 			sent := false
 			r.RLock()
 			m := messageFromPublish(message, ackFunc(ackInChan, client.persist, message))
-			var handlers []MessageHandler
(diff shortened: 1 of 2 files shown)


alespour:
I cannot sign into Eclipse and sign ECA. If anyone finds this PR useful, please use the commits anyway suitable for your own PR. Thanks.

MattBrittan:
Sorry for the long delay in responding to this, as you say it's a very subtile optimization but, I guess, the router forms part of the hot path so it's worth the slight decrease in readability. Thanks for the PR.

PR #713, opened 2025-05-23
Optimize TCP connection logic

geekeryy:
Optimize TCP connection logic, use proxy settings in the environment instead of the original all_proxy environment variable check
Supports both ALL_PROXY and all_proxy environment variables


MattBrittan:
`FromEnvironmentUsing` was added in 2019 (this code was written prior to that), so this seems like a sensible change - thanks.

PR #719, opened 2025-07-10
Update filestore.go

AlexTransit:
removes a panic if there is no rights to a file with an outgoing message.

### DIFF
--- a/filestore.go
+++ b/filestore.go
@@ -67,7 +67,7 @@ func (store *FileStore) Open() {
 
 	// if store dir exists, great, otherwise, create it
 	if !exists(store.directory) {
-		perms := os.FileMode(0770)
+		perms := os.FileMode(0o770)
 		merr := os.MkdirAll(store.directory, perms)
 		chkerr(merr)
 	}
@@ -113,14 +113,16 @@ func (store *FileStore) Get(key string) packets.ControlPacket {
 		return nil
 	}
 	mfile, oerr := os.Open(filepath)
-	chkerr(oerr)
+	if oerr != nil {
+		ERROR.Println(STR, "file openning error:", oerr)
+	}
 	msg, rerr := packets.ReadPacket(mfile)
-	chkerr(mfile.Close())
+	mfile.Close()
 
 	// Message was unreadable, return nil
 	if rerr != nil {
 		newpath := corruptpath(store.directory, key)
-		WARN.Println(STR, "corrupted file detected:", rerr.Error(), "archived at:", newpath)
+		WARN.Println(STR, "corrupted file detected:", rerr.Error(), " archived at:", newpath)
 		if err := os.Rename(filepath, newpath); err != nil {
 			ERROR.Println(STR, err)
 		}


MattBrittan:
Thanks - would you mind raising an issue re this? Whilst at first glance this appears a simple fix I think some discussion might be appropriate, if only so future viewers can understand why the change was made  (there are reasons the current approach was selected).

I don't think there is a "right" option here, you can argue either way:

* Being unable to load the file could cause data loss, so `panic` is appropriate (forces operator to address the issue). It seems likely that this will generally be an all or nothing issue (i.e. can't read ANY files from the folder) and generally occur the first time the app runs (because there is an issue with the setup).
* In some cases missing a message is better than panicing (i.e. issue impacts one file, code is running on a remote unit)

Either way if failing to read a file is no longer a `panic` should the same apply to writing a file? (you can argue that failure to write is more likely to lead to data loss but often the data is time critical..).

The [v5 client](https://github.com/eclipse-paho/paho.golang/blob/1db71fa3ccf05172a1a7850a8bc6d08cd7679d84/paho/store/file/store.go#L123) does just log this specific error. However, in order to fail fast, it [checks that it can write/read]((https://github.com/eclipse-paho/paho.golang/blob/1db71fa3ccf05172a1a7850a8bc6d08cd7679d84/paho/store/file/store.go#L59) ) from the folder. I prefer this approach because otherwise it's easy to miss a configuration issue that prevents QOS1+ operating correctly. Unfortunately we are constrained by a desire to avoid breaking changes in this client (so maybe a `panic` in `NewFileStore` if a test write/read fails?)...

Note that after `ERROR.Println(STR, "file openning error:", oerr)` you should proably should `return nil` (also note the spelling error).

Note: I can't accept the PR without a signed [ECA](https://www.eclipse.org/legal/eca/)

MattBrittan:
> It is better to have an error in the log than a non working program.

I think that depends upon the situation. As apps running MQTT are often long running, run on remote kit,  and are fairly much unmonitored, logged messages are often ignored (I suspect that quite a few users will not even enable logging due to the way it works in this library). This means that an immediate panic (i.e. when the app is deployed and the issue is easily noticed/fixed) can be preferable to an ongoing, unnoticed problem.

This is what I suggested the technique used in the V5 client which, I feel, provides the best of both (the app either panics at startup if there is an issue with the storage, or continues to run if the issue occurs later). It would also be possible to provide an additional `NewFileStoreEx` (or similar) that returns the error instead of panicing.

> The work of QOS1+ is not violated. The message is saved.

Message order is important in MQTT, so not retransmitting the message is an issue. Obviouslly how important this is depends on your use-case.

I suspect we are considering different things here. Your issue was a one-off (file written when app running as root, failed to read on restart) whereas I'm more concerned about longer term issues (reads fail on every packet) which could easily lead to ongoing message loss (each time the connection drops a few messages are lost). I have had issues before where I've been happy to see the `Panic` because it highlighted a configuration issue that could be fixed before the unit was shipped.

Note that I'm not against the change, just think it's important to consider the implications - hence the request that you raise an issue (as I said there are reasons the code works as it does).

MattBrittan:
Have raised issue #720 so this is not missed. Whilst I don't think removing all of the panics is appropriate it would be good to improve the situation.

AlexTransit:
> Have raised issue #720 so this is not missed
Thanks

PR #735, opened 2025-11-11
fix: remove unnecessary if statement

mojixcoder:
`TrimPrefix` returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged.

Here is the function definition:
```go
func TrimPrefix(s, prefix string) string {
	if HasPrefix(s, prefix) {
		return s[len(prefix):]
	}
	return s
}
```


MattBrittan:
Thanks, but cannot accept this without an ECA (see the check above, or readme, for more details, yell out if you are having issues signing).

PR #737, opened 2025-12-01
Check for invalid "Remaining Length" when encoding/decoding

MattBrittan:
The check when decoding was added to the MQTT spec in errata01 (in 2015). Makes sense to also check when encoding to avoid sending an invalid packet.


MattBrittan:
closes #739 

(issue raised after discussion of security implications of this bug).

PR #748, opened 2026-03-04
Fix typos in comments for clarity and consistency

abdirizak-alaaja:
No functional changes were made. The updates are limited to comment text and do not affect behavior.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,7 @@
  *    Allan Stockdill-Mander
  *    Mike Robertson
  *    Matt Brittan
+ *	  Abdirizak Abdullahi
  */
 
 // Portions copyright © 2018 TIBCO Software Inc.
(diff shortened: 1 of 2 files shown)


MattBrittan:
Thanks for your contribution. Unfortunately I cannot accept it without a signed ECA.

Note that checking/accepting PR's like this can take longer than it takes to create the PR, so I'd appreciate it if you could fully review a file before submitting if possible (e.g. "finnish" is used twice in `client.go` but you corrected it once). QOS is a well known term in MQTT so I don't think the expansion is necessary (particularly in lower level code like `PingreqPacket`).

abdirizak-alaaja:
alright man

MattBrittan:
Closing due to no ECA (have implemented many of the changes in an additional PR).

PR #750, opened 2026-05-23
Suppress spurious "BUG BUG BUG reconnection function is nil" log when AutoReconnect disabled

### LINKED ISSUE #697: disabling AutoReconnect causes spurious error to be logged
We set `AutoReconnect` to false in our `ClientOptions`. When the connection to the broker is lost unexpectedly, we see the following error come from this library:
> [client]  BUG BUG BUG reconnection function is nil<nil>

Tracing through the code, this is because `disDone` is assigned like so.
https://github.com/eclipse-paho/paho.mqtt.golang/blob/714f7c0231294ec4144f0e0e5fc5b43a6d430d2f/client.go#L517

Since `c.options.AutoReconnect` is false, that will pass false to `c.status.ConnectionLost`. Thus the callback that it returns will itself return `nil, nil`.
https://github.com/eclipse-paho/paho.mqtt.golang/blob/714f7c0231294ec4144f0e0e5fc5b43a6d430d2f/status.go#L278-L280

This in turn triggers the log that "should never happen".
https://github.com/eclipse-paho/paho.mqtt.golang/blob/714f7c0231294ec4144f0e0e5fc5b43a6d430d2f/client.go#L547-L552

I believe this regression was introduced in v1.4.2. It prevents us from upgrading from v1.3.5.

ultramcu:
## Summary

Fixes #697.

When `AutoReconnect` is disabled and the connection drops, the library logs a misleading error:

```
[client]  BUG BUG BUG reconnection function is nil<nil>
```

This is harmless but alarming, and it has prevented users from upgrading past v1.3.5.

## Root cause

`internalConnLost` calls `c.status.ConnectionLost(c.options.AutoReconnect && c.status.ConnectionStatus() > connecting)`. When `AutoReconnect` is `false`, `willReconnect` is `false`, so the handler returned by `getConnectionLostHandler` legitimately returns `(nil, nil)` (see `status.go` — `if !reconnectRequested || !proceed { return nil, nil }`).

Back in `internalConnLost`, the completion handler then logs `BUG BUG BUG reconnection function is nil` whenever `reConnDone == nil`, without checking whether a reconnect was ever expected. A `nil` reconnect function is the *expected* outcome when no reconnect was requested.

(Tracing the state machine: `reConnDone == nil && err == nil` only occurs when `reconnectRequested == false`; if a reconnect was requested but cancelled in the interim, `disDone(true)` returns a non-nil error instead, which is already handled by the preceding branch. So the `BUG` branch was unreachable except in this expected, non-buggy case.)

## Fix

As suggested by @MattBrittan in the issue, capture whether a reconnect is expected and only emit the log in that case:

```go
reconnectExpected := c.options.AutoReconnect && c.status.ConnectionStatus() > connecting
disDone, err := c.status.ConnectionLost(reconnectExpected)
...
} else if reConnDone == nil && reconnectExpected { // Should never happen
    c.logger.Error("BUG BUG BUG reconnection function is nil", ...)
}
```

This preserves the genuine "should never happen" guard for the reconnect-expected path while suppressing the spurious log when reconnect was never requested. Behaviour is unchanged when `AutoReconnect` is enabled.

## Testing

- Added `Test_internalConnLost_NoBugLogWhenAutoReconnectDisabled` (white-box), which drives `internalConnLost` with `AutoReconnect=false` and asserts the `BUG BUG BUG` log is not emitted. Verified it **fails without** the fix and **passes with** it.
- Ran the full broker-free unit suite with `-race` — all pass.
- Ran the broker-dependent tests that exercise `internalConnLost` (`Test_autoreconnect`, `Test_cleanUpMids`, `Test_ResumeSubsWithReconnect`) against a local mosquitto broker — all pass.

### DIFF
--- a/client.go
+++ b/client.go
@@ -561,7 +561,11 @@ func (c *client) internalConnLost(whyConnLost error) {
 	// (including after sending a DisconnectPacket) as such we only do cleanup etc if the
 	// routines were actually running and are not being disconnected at users request
 	c.logger.Debug("internalConnLost called", slog.String("component", string(CLI)))
-	disDone, err := c.status.ConnectionLost(c.options.AutoReconnect && c.status.ConnectionStatus() > connecting)
+	// reconnectExpected indicates whether we anticipate transitioning into a reconnect. When AutoReconnect is
+	// disabled (or we were not fully connected) the connection lost handler will return a nil reconnect function
+	// and that is the expected behaviour (rather than a bug).
+	reconnectExpected := c.options.AutoReconnect && c.status.ConnectionStatus() > connecting
+	disDone, err := c.status.ConnectionLost(reconnectExpected)
 	if err != nil {
 		if err == errConnLossWhileDisconnecting || err == errAlreadyHandlingConnectionLoss {
 			return // Loss of connection is expected or already being handled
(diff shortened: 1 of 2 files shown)


MattBrittan:
Looks good to me - thanks for resolving this.

Issue #720, opened 2025-07-11
filestore - panic when file cannot be read


MattBrittan:
This issue was triggeded by the now closed PR #719 as I feel this is something worth discussing/changing.

Currently the filestore panics in a number of situations:

- `Open` if the folder does not exist and can not be created
- If a file cannot be written
- If a file cannot be read (but not if the data is read successfully but is corrupted)

One situation where this can lead to a panic is (copied from the PR):

1. The program was launched with Root rights when the broker was unavailable.
1. The message was saved for subsequent sending. With Root rights.
1.  When the program was launched with limited rights, we get a panic.

I believe it can be argued that a panic is appropriate because a non-working file system can lead to a loss of messages. However it can also be argued that a panic is excessive and the system should log the issue and resume.

Personally I'm reluctant to remove all of the panics. If there is an issue that impacts message integrity then panicing (in the absence of an option to log an error) is probably appropriate - especially soon after startup (as this means its more likely the user will spot/correct the issue). The V5 client takes a different approach, it checks that it can read/write from the folder at startup and then just logs any errors after that (felt this was a pragmatic solution). Unfortunately the current V3 client API does not have provision to return errors.

Thoughts?

PR #751, opened 2026-05-23
filestore: fail fast at startup and stop panicking on unreadable files

ultramcu:
## Summary

Addresses #720 — the `FileStore` panics in several situations where a file cannot be read/written. This implements the V5-client approach you outlined in #720 / #719.

## Behaviour change

**1. Fail fast at startup.** `Open()` now verifies the store directory can be written to, read from, and have files removed (by writing/reading/removing a small temporary file). If any step fails it panics, surfacing configuration problems (most commonly insufficient permissions) at startup — when an operator is most likely to notice and fix them — rather than later as a panic mid-operation or as silent message loss on a long-running, unattended client. This mirrors [the v5 client](https://github.com/eclipse-paho/paho.golang/blob/1db71fa3ccf05172a1a7850a8bc6d08cd7679d84/paho/store/file/store.go#L59). The `Store` interface doesn't allow `Open()` to return an error, so (as you suggested) a `panic` is used.

**2. Don't panic on an unreadable individual file.** Once the store is open, `Get()` now archives a file it cannot open as `.CORRUPT` and returns `nil` (consistent with the existing handling for corrupt *contents*), instead of panicking. This fixes the original scenario from #719: a message written while running as `root`, then read after restarting with reduced privileges, no longer crashes the client.

## Deliberately left as-is (open to your call)

- **Write failures still panic.** Per your comments in #720, failing to *write* is more likely to indicate data loss, and the new startup check already catches the common "directory not writable" misconfiguration. I kept `write()` panicking rather than widening the scope.
- I did **not** change the `Store` interface or add a `NewFileStoreEx`-style error-returning constructor, to avoid a breaking change. Happy to add one if you'd prefer that route.

## One implication worth noting

A store directory that is **readable but not writable** (e.g. read-only, containing messages only to be read) previously let `Open()` and `Get()` work and only panicked on the first `Put()`. It now panics at `Open()` via the startup check. This is consistent with the "fail fast on a config issue that breaks QOS1+" intent, but flagging it explicitly in case that edge case matters to anyone.

## Testing

- Added `Test_FileStore_Open_FailsFastWhenDirectoryUnusable` — points the store at a path that is a regular file so the startup write-test cannot succeed; asserts `Open()` panics. (Uses a regular file rather than permissions so it's independent of the user it runs as.)
- Added `Test_FileStore_Get_UnreadableFileArchived` — `chmod 0000` on a stored file, asserts `Get()` does **not** panic, returns `nil`, and archives the file as `.CORRUPT`. (Skipped on Windows / when running as root, where file permissions don't apply.)
- Both new tests were confirmed to **fail without** this change and **pass with** it.
- Full broker-free suite (incl. all store tests) passes with `-race`.

### DIFF
--- a/filestore.go
+++ b/filestore.go
@@ -19,6 +19,7 @@
 package mqtt
 
 import (
+	"fmt"
 	"io/fs"
 	"log/slog"
 	"os"
@@ -33,6 +34,10 @@ const (
 	msgExt     = ".msg"
 	tmpExt     = ".tmp"
 	corruptExt = ".CORRUPT"
+	// writeTestFile is the name of the temporary file written, read back and removed by Open()
+	// to confirm the store directory is usable. The leading dot and reserved name avoid any
+	// clash with real message files (whose keys have the form "i.<id>"/"o.<id>").
+	writeTestFile = ".paho-write-test" + tmpExt
 )
 
 // FileStore implements the store interface using the filesystem to provide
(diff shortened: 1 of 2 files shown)


MattBrittan:
I think this is a decent compromise (did not really make sense to panic if `Open` failed but not if the file was corrupt). Thanks.

PR #752, opened 2026-05-24
client: add public DeleteRoute to remove a route added with AddRoute

### LINKED ISSUE #331: DeleteRoute implementation
Looking at the Client API and codebase, it is possible to add a route with a callback, but impossible to delete a route. The code is there - `deleteRoute` in router.go and it's called from Unsubscribe but there's no way to get at it from the public API.

ultramcu:
## Summary

Closes #331.

`Client.AddRoute` lets you register a message handler for a topic, but there was no public way to remove one again. The internal `router.deleteRoute` already exists (and is used by `Unsubscribe`), it just wasn't reachable through the `Client` API.

This adds `DeleteRoute(topic string)` to the `Client` interface and the `client` implementation, mirroring `AddRoute`. It simply calls `router.deleteRoute`, which already takes the router lock, so it is safe for concurrent use (per the note on the issue). Deleting a topic with no registered handler is a no-op.

```go
c.AddRoute("sensors/+/temp", handler)
// ...
c.DeleteRoute("sensors/+/temp") // remove the handler again
```

Note: this only removes the local route handler; it does not unsubscribe (use `Unsubscribe` for that), consistent with `AddRoute` not subscribing.

## Testing

Added `Test_Client_AddDeleteRoute`: registers a route via `Client.AddRoute`, asserts it is present, removes it via `DeleteRoute`, asserts it is gone, and checks that deleting an unknown topic is a no-op (no panic).

- `go test -race` (full broker-free unit suite) passes; `go vet` and `gofmt` clean.
- Confirmed no other types implement the `Client` interface, so adding the method does not break any existing code.

### DIFF
--- a/client.go
+++ b/client.go
@@ -114,6 +114,10 @@ type Client interface {
 	// a new go routine.
 	// callback must be safe for concurrent use by multiple goroutines.
 	AddRoute(topic string, callback MessageHandler)
+	// DeleteRoute removes the handler previously added for the given topic with
+	// AddRoute. It is a no-op if no handler is registered for that exact topic.
+	// Note that this does not unsubscribe; use Unsubscribe for that.
+	DeleteRoute(topic string)
 	// OptionsReader returns a ClientOptionsReader which is a copy of the clientoptions
 	// in use by the client.
 	OptionsReader() ClientOptionsReader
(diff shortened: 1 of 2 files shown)


MattBrittan:
Thanks for this - looks good. I had been concerned that using this in a callback could result in a deadlock but this does not appear to be the case (`matchAndDispatch` either calls the handler in a go routine, or releases the lock before doing so).
It might be worth adding a note that it is possible for a handler to be called after `DeleteRoute` returns (due to the use of go routines there are no guarantees) but that seems pretty minor.

PR #797, opened 2026-09-22
Reject invalid SUBACK return codes

### LINKED ISSUE #761: SUBACK accepts reserved Granted QoS bits
## Standard Requirement

MQTT 3.0, `3.9 SUBACK - Subscription acknowledgement`, Payload:

Reference: [MQTT v3.0/v3.1 specification, Section 3.9, SUBACK - Subscription acknowledgement](https://public.dhe.ibm.com/software/dw/webservices/ws-mqtt/mqtt-v3r1.html).

```text
The table below shows the Granted QoS field encoded in a byte.

bit    7 6 5 4 3 2 1 0
       Reserved Reserved Reserved Reserved Reserved Reserved QoS level

The upper 6 bits of this byte are not used in the current version of the protocol. They are reserved for future use.
```

Only bits 1 and 0 define the Granted QoS value; bits 7 through 2 are reserved.

## Relevant Source Code

`implementions/paho.mqtt.golang-master/packets/suback.go:37`

```go
func (sa *SubackPacket) Write(w io.Writer) error {
	var body bytes.Buffer
	body.Write(encodeUint16(sa.MessageID))
	body.Write(sa.ReturnCodes)
```

`implementions/paho.mqtt.golang-master/packets/suback.go:55`

```go
func (sa *SubackPacket) Unpack(b io.Reader) error {
	var qosBuffer bytes.Buffer
	sa.MessageID, err = decodeUint16(b)
	_, err = qosBuffer.ReadFrom(b)
	sa.ReturnCodes = qosBuffer.Bytes()
	return nil
}
```

`implementions/paho.mqtt.golang-master/net.go:221`

```go
if t, ok := token.(*SubscribeToken); ok {
	if len(m.ReturnCodes) != len(t.subs) {
		token.setError(ErrMalformedSuback)
		continue
	}
	for i, qos := range m.ReturnCodes {
		t.subResult[t.subs[i]] = qos
	}
}
```

`implementions/paho.mqtt.golang-master/token.go:188`

```go
func (s *SubscribeToken) Result() map[string]byte {
	s.m.RLock()
	defer s.m.RUnlock()
	return s.subResult
}
```

## Implementation Behavior

`SubackPacket.Unpack` stores Granted QoS bytes unchanged. The client-side SUBACK handler validates only the number of return codes, then copies each raw byte into `SubscribeToken.subResult`. No code masks or rejects reserved upper bits.

## Inconsistency Reason

The standard reserves bits 7 through 2 of each Granted QoS byte. Paho gives those bits current observable meaning by accepting `0xfe` and returning it to the application as the granted subscription result.

## Runtime Evidence

Packet parser probe:

```text
..\..\..\..\..\tools\go\bin\go.exe test . -run "^Test(PositiveControlSubackGrantedQos2|SubackGrantedQosUpperBitsAccepted)$" -count=1 -v
```

Output:

```text
=== RUN   TestPositiveControlSubackGrantedQos2
--- PASS: TestPositiveControlSubackGrantedQos2 (0.00s)
=== RUN   TestSubackGrantedQosUpperBitsAccepted
--- PASS: TestSubackGrantedQosUpperBitsAccepted (0.00s)
PASS
ok  	speclitmus-mqtt30-paho-501-650-packet-probe	0.568s
```

The client path probe then connected to a test broker that returned either a valid Granted QoS byte or a malformed byte with reserved upper bits set:

```text
..\..\..\..\..\tools\go\bin\go.exe test . -run "^TestClient.*Suback" -count=1 -v
```

Output:

```text
=== RUN   TestClientAcceptsValidSubackGrantedQoS2
    client_suback_granted_qos_test.go:90: SUBACK granted byte=0x02 result=map[string]uint8{"topic/a":0x2}
--- PASS: TestClientAcceptsValidSubackGrantedQoS2 (0.01s)
=== RUN   TestClientExposesSubackGrantedQoSUpperBits
    client_suback_granted_qos_test.go:97: SUBACK granted byte=0xfe result=map[string]uint8{"topic/a":0xfe}
--- PASS: TestClientExposesSubackGrantedQoSUpperBits (0.00s)
PASS
ok  	speclitmus-mqtt30-paho-client-suback-granted-qos-probe	0.050s
```

## Impact

A broker can send a malformed MQTT 3.0 SUBACK Granted QoS byte such as `0xfe`; Paho treats it as successful and exposes the reserved bits to application code.

## Fix Direction

Reject malformed SUBACK Granted QoS bytes whose value is not a valid MQTT 3.0 granted QoS, or at minimum reject any byte with `qos & 0xfc != 0`.

ajavanma:
## Before submitting

This is a focused fix for the existing SUBACK validation issue.

## Related issue

Closes #761

## Description

A broker can return a reserved SUBACK value such as `0xfe`, and the client currently completes the subscription successfully with that value in `SubscribeToken.Result()`.

Validate all return-code values in `SubackPacket.Unpack` before assigning `ReturnCodes`. Accept QoS 0, 1, and 2, plus `0x80` (subscription refused), for every supported protocol version. Accepting `0x80` for MQTT 3.1 preserves compatibility with brokers that use it for both MQTT 3.1 and 3.1.1.

Invalid values produce a decoding error and follow the existing connection-loss path without publishing partial subscription results. The existing `mqtt.ErrMalformedSuback` aliases the packet error so callers can recognize it with `errors.Is`. Return-code count validation remains in the subscription handler, where the original request is available. The protocol-version accessor is no longer needed, and existing subscription-resumption behavior is preserved.

## Testing

- Added 24 decoder cases covering valid/refused values, reserved values, and invalid values after a valid prefix through both `Unpack` and `ReadPacket`; the following packet remains readable.
- Added 10 full-client `net.Pipe` cases across MQTT 3.1, MQTT 3.1.1, both bridge modes, and automatic fallback. Valid refusals keep the connection open; malformed values close it, leave results empty, and complete/free the pending subscription under the default cleanup policy.
- Retained and expanded nine handler cases for valid results and return-code count mismatches.
- Before this revision, 14 decoder cases and three MQTT 3.1 refusal compatibility cases failed; all pass after the revision.
- Passed `TEST_FVT_ADDR=127.0.0.2 go test -race ./... -count=1 -timeout=10m` against an isolated Mosquitto broker on ARM64, plus `go vet ./...`, formatting, and diff checks.


net.go
```diff
@@ -217,16 +217,27 @@ func startIncomingComms(conn io.Reader,
...
 				logger.Debug("startIncomingComms: received suback", slog.Uint64("messageID", uint64(m.MessageID)), slog.String("component", string(NET)))
 				token := c.getToken(m.MessageID)
+				t, isSubscribe := token.(*SubscribeToken)
+				// [MQTT-3.8.4-5] requires one return code per requested subscription.
+				malformed := isSubscribe && len(m.ReturnCodes) != len(t.subs)
+				protocolVersion := c.getProtocolVersion()
+				for _, qos := range m.ReturnCodes {
+					// MQTT 3.1 permits only QoS 0, 1 and 2. MQTT 3.1.1 also
+					// permits 0x80 (failure), including in bridge mode.
+					if qos > 2 && !(qos == 0x80 && (protocolVersion == 4 || protocolVersion == 0x84)) {
```
MattBrittan: Whilst this is technically correct against the MQTT 3.1 specification, I think we should accept 0x80 regardless of the protocol version.

Some brokers share SUBACK handling between MQTT 3.1 and 3.1.1 and can return 0x80 for either version. Examples include [VerneMQ](https://github.com/vernemq/vernemq/blob/main/apps/vmq_server/src/vmq_mqtt_fsm.erl#L460-L507) and [EMQX](https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_channel.erl#L1558-L1562). Rejecting these responses could break live code where these brokers are in use and the protocol version is set to 3.1 (no idea if anyone does this but..). Mosquitto handles this differently and avoids 0x80 for MQTT 3.1 authorization failures.

This is similar to the acceptance of `protocolVersion == 0x84`. Technically this violates the spec, but was  included because some brokers use it to indicate bridge mode (an unofficial extension).  I feel we should take the same approach here - accept 0x80 to maintain compatibility (there seems little real benefit to enforcing this more stringently).

If we accept 0x80 across versions, would it make sense to validate the return-code values during decoding in `packets/suback.go`? The check would no longer need the negotiated protocol version. The return-code count check would still belong in the subscription handler, where the original request is available. I'm raising this because we are seeing a lot more issues around validating received packets (suspect AI code reviews) and it probably makes sense to implement such checks when decoding (so protocol violation is detected before any other code relies upon the result). I note that there is an unused `Validate` function in `packets/connect.go` 

Examples [VerneMQ](https://github.com/vernemq/vernemq/blob/main/apps/vmq_server/src/vmq_mqtt_fsm.erl#L460-L507), [EMQX](https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_channel.erl#L1558-L1562). Mosquitto does have a seperate path for 3.1 so will not return this.

MattBrittan (changes requested):
Sorry - your PR addresses the issue perfectly. However I feel that we should probably allow some deviation from the spec to better support existing brokers (don't want to break existing usage).

net.go
ajavanma: Addressed in e450e07. `0x80` is now accepted across all supported protocol versions, and value validation happens in `SubackPacket.Unpack`. The count check remains in the subscription handler. Added decoder and client regressions; the full race suite and `go vet` pass.

MattBrittan: Thanks—this addresses my concerns. Looks good to me.

